PR #35, opened 2024-07-23
PIM implementation of transitive closure

hugoabbot:
*Algorithm needs further reviewing to ensure that it's optimal and valid. As discussed in weekly meeting, this benchmark might be added before the final submission deadline if time permits. 

*Needs to be moved from PIMbench to misc-bench

## Transitive Closure

Benchmark uses the Floyd-Warhsall algorithm to perform transitive closure on a directed and weighted graph. `loadVector()` is not used in the timing as its functionality is possible using just PIM commands, but is not currently supported by the current API.

## Input

Specified input is supported given that the format is valid and supported. Examples are given in `datafiles/`. Additionally, the executable can be run without this specification, which defaults to a matrix with `256` nodes and a sparsity rate of `50%`.


PIMbench/cpp-transitive-closure/Makefile
```diff
@@ -0,0 +1,17 @@
+# Makefile: C++ version of transitive closure
+# Copyright 2024 LavaLab @ University of Virginia. All rights reserved.
```
deyuan, line 2: This copyright info needs to be updated

PIMbench/cpp-transitive-closure/Makefile
```diff
@@ -0,0 +1,17 @@
...
+
+include ../Makefile.common
+
+EXEC := transitive-closure.out
+SRC := transitive-closure.cpp
+
+FOPENMP := -fopenmp
```
deyuan, line 9: Need to guard this with USE_OPENMP flag. Example: cpp-gemm/Makefile

fasiddique:
I would like to keep this on hold for now. If time permits we can add it to the paper.

fasiddique (approved):
Thank you!

PR #53, opened 2024-07-24
Updated default parameters for all the PIMBench to be a smaller value. **No Significant code change. Please review and merge asap**

fasiddique:
Tested all the modified benchmark with the default values. This change has been done so that AE is faster.


PIMbench/cpp-gemm/gemm.cpp
```diff
@@ -207,7 +207,7 @@ void gemm(uint64_t row, uint64_t colA, uint64_t colB, std::vector<std::vector<in
 int main(int argc, char *argv[])
 {
   struct Params params = getInputParams(argc, argv);
-  std::cout << "Row: " << params.row << " Column: " << params.columnA << "\n";
+  std::cout << "Running gemm on PIM for matrix row: " << params.row << " column: " << params.columnA << "\n";
```
KumareshN93, line 210: I think we can also print params.columnB here since columnA and columnB are different by default.

KumareshN93 (approved):
Changes look good.

fasiddique:
> Changes look good.

Thank you for reviewing and helpful suggestion!

PR #57, opened 2024-07-25
Minor changes to linear regression (mostly code cleanup) and util function now generates vector in range 0-1024

fasiddique:


PIMbench/util.h
```diff
@@ -21,14 +21,16 @@
...
+
 void getVector(uint64_t vectorLength, std::vector<int> &srcVector)
 {
   srand((unsigned)time(NULL));
   srcVector.resize(vectorLength);
 #pragma omp parallel for
   for (uint64_t i = 0; i < vectorLength; ++i)
   {
-    srcVector[i] = (rand() % (i + 1) + 1);
+    srcVector[i] = (rand() % MAX_NUMBER);
```
KumareshN93, line 33: Do we also need to update getMatrix() function  as above..

PIMbench/cpp-linear-regression/lr.cpp
```diff
@@ -203,34 +174,34 @@ int main(int argc, char *argv[])
...
+  auto intercept_device = (double)(SY_device - slope_device * SX_device) / params.dataSize;
   auto end = std::chrono::high_resolution_clock::now();
   hostElapsedTime += (end - start);
 
   if (params.shouldVerify)
   {
     // verify result
-    int SX = 0, SY = 0, SXX = 0, SYY = 0, SXY = 0;
+    int64_t SX = 0, SY = 0, SXX = 0, SXY = 0;
 #pragma omp parallel for reduction(+ : SX, SXX, SY, SYY, SXY)
```
KumareshN93, line 192: Should we remove SYY from the pragma?

PIMbench/cpp-linear-regression/lr.cpp
```diff
@@ -203,34 +174,34 @@ int main(int argc, char *argv[])
...
       SXY += dataPointsX[i] * dataPointsY[i];
     }
+
     // Calculate slope and intercept
-    auto slope = (params.dataSize * SXY - SX * SY) / (params.dataSize * SXX - SX * SX);
-    auto intercept = (SY - slope * SX) / params.dataSize;
-    if (intercept != intercept_device)
+    auto slope = (double)(params.dataSize * SXY - SX * SY) / (params.dataSize * SXX - SX * SX);
+    auto intercept = (double)(SY - slope * SX) / params.dataSize;
+    if ((int)intercept != intercept_device)
```
KumareshN93, line 204: Quick question. intercept is typecast to int while intercept_device is double. Is this expected?

KumareshN93 (approved):
Changes look good.

PIMbench/util.h
fasiddique: Updated thanks!

PR #61, opened 2024-07-25
Parse Config File

mhgholamrezaei:
Changes:
- Add utility functions to the Utils module
- Added config file for DRAM
- Parse the config file to get the DRAM dimensions for functional simulation in pimDevice
- Parse the config file to get the DRAM all parameters for the performance model
- Get the config files path and names from the envirnment variables: 

According to the specification in #126 in old repo.

### DIFF
--- a/configs/PIMeval_BitSimdV.cfg
+++ b/configs/PIMeval_BitSimdV.cfg
@@ -4,5 +4,5 @@ num_subarray_per_bank = 32
 num_row_per_subarray = 1024
 num_col_per_subarray = 8192
 simulation_target = PIM_DEVICE_BITSIMD_V
-max_num_threads = 10
+max_num_threads = 10; Maximum Number of Threads
 memory_config_file = DDR4_8Gb_x16_3200.ini
--- a/configs/PIMeval_BitSimdV.cfg
+++ b/configs/PIMeval_BitSimdV.cfg
@@ -1,0 +1,8 @@
+num_ranks = 2
+num_bank_per_rank = 128
+num_subarray_per_bank = 32
+num_row_per_subarray = 1024
+num_col_per_subarray = 8192
+simulation_target = PIM_DEVICE_BITSIMD_V
+max_num_threads = 10
+memory_config_file = DDR4_8Gb_x16_3200.ini
--- a/libpimeval/src/libpimeval.h
+++ b/libpimeval/src/libpimeval.h
@@ -30,6 +30,7 @@ enum PimDeviceEnum {
   PIM_DEVICE_BITSIMD_H,
   PIM_DEVICE_FULCRUM,
   PIM_DEVICE_BANK_LEVEL,
+  PIM_DEVICE_UNKNOWN,
 };
 
 //! @brief  PIM allocation types
--- a/libpimeval/src/libpimeval.h
+++ b/libpimeval/src/libpimeval.h
@@ -30,7 +30,6 @@ enum PimDeviceEnum {
   PIM_DEVICE_BITSIMD_H,
   PIM_DEVICE_FULCRUM,
   PIM_DEVICE_BANK_LEVEL,
-  PIM_DEVICE_UNKNOWN,
 };
 
 //! @brief  PIM allocation types
--- a/libpimeval/src/pimDevice.cpp
+++ b/libpimeval/src/pimDevice.cpp
@@ -301,6 +301,7 @@ pimDevice::parseConfigFromFile(const std::string& config, unsigned& numRanks, un
   std::unordered_map<std::string, std::string> params;
 
   while (std::getline(configStream, line)) {
+    line = pimUtils::removeAfterSemicolon(line);
     if (line.empty() || line[0] == '[') {
       continue;
     }
--- a/libpimeval/src/pimDevice.cpp
+++ b/libpimeval/src/pimDevice.cpp
@@ -69,26 +69,21 @@ pimDevice::adjustConfigForSimTarget(unsigned& numRanks, unsigned& numBankPerRank
   return true;
 }
 
-//! @brief  Config device type and simulation target
+//! @brief  Config the simulation target
 void
-pimDevice::configDevice(PimDeviceEnum curDevice, PimDeviceEnum simTarget)
+pimDevice::configSimTarget(PimDeviceEnum simTarget)
 {
-  m_deviceType = curDevice;
-  m_simTarget = curDevice;
+  m_simTarget = simTarget;
   // from 'make PIM_SIM_TARGET=...'
   #if defined(PIM_SIM_TARGET)
-  if (simTarget == PIM_DEVICE_NONE) {
-    simTarget = PIM_SIM_TARGET;
+  if (m_simTarget == PIM_DEVICE_NONE) {
+    m_simTarget = PIM_SIM_TARGET;
   }
   #endif
 
-  // determine simulation target for functional device
-  if (curDevice == PIM_FUNCTIONAL) {
-    // default sim target
-    if (simTarget == PIM_DEVICE_NONE || simTarget == PIM_FUNCTIONAL) {
-      simTarget = PIM_DEVICE_BITSIMD_V;
-    }
-    m_simTarget = simTarget;
+  // Default simulation target
+  if (m_simTarget == PIM_DEVICE_NONE) {
+    m_simTarget = PIM_DEVICE_BITSIMD_V;
   }
 }
 
@@ -150,21 +145,30 @@ bool
 pimDevice::init(PimDeviceEnum deviceType, unsigned numRanks, unsigned numBankPerRank, unsigned numSubarrayPerBank, unsigned numRows, unsigned numCols)
 {
   assert(!m_isInit);
-  assert(deviceType != PIM_DEVICE_NONE);
 
-  std::printf("PIM-Info: Trying to read it from envirnment variable %s\n", pimUtils::envVarPimEvalTarget);
-  // Read envirnment variable for the simulation target
-  std::string pimEvalTarget;
-  if (!pimUtils::getEnvVar(pimUtils::envVarPimEvalTarget, pimEvalTarget)) {
-    std::printf("PIM-Info: Could not read environment variable %s\n", pimUtils::envVarPimEvalTarget);
+  // Determine simulation target
+  m_simTarget = deviceType;
+  if (m_simTarget == PIM_FUNCTIONAL) {
+    m_simTarget = PIM_DEVICE_BITSIMD_V;
   }
-  m_simTarget = pimUtils::strToPimDeviceEnum(pimEvalTarget);
-  if (m_simTarget == PIM_DEVICE_NONE) {
-    std::printf("PIM-Error: Invalid simulation target %s\n", pimEvalTarget.c_str());
-    assert (true);
-  }
-  if (m_simTarget == PIM_DEVICE_NONE) {
-    configDevice(deviceType);
+  if (deviceType == PIM_DEVICE_NONE) {
+    // Read envirnment variable for the simulation target
+    bool readSimTargetFromMakeArgument = false;
+    std::printf("PIM-Info: Trying to read simulation target from envirnment variable %s\n", pimUtils::envVarPimEvalTarget);
+    std::string pimEvalTarget;
+    bool readEnvVarStatus = pimUtils::getEnvVar(pimUtils::envVarPimEvalTarget, pimEvalTarget);
+    if (!readEnvVarStatus) {
+      std::printf("PIM-Info: Could not read environment variable %s\n", pimUtils::envVarPimEvalTarget);
+      readSimTargetFromMakeArgument = true;
+    }
+    m_simTarget = pimUtils::strToPimDeviceEnum(pimEvalTarget);
+    if (m_simTarget == PIM_DEVICE_NONE) {
+      std::printf("PIM-Warning: Invalid value %s for environment varialbe %s\n", pimEvalTarget.c_str(), pimUtils::envVarPimEvalTarget);
+      readSimTargetFromMakeArgument = true;
+    }
+    if (readSimTargetFromMakeArgument){
+      configSimTarget(m_simTarget);
+    }
   }
   std::printf("PIM-Info: Current Device = %s, Simulation Target = %s\n",
               pimUtils::pimDeviceEnumToStr(m_deviceType).c_str(),
@@ -209,8 +213,6 @@ pimDevice::init(PimDeviceEnum deviceType, const char* configFileName)
 {
   bool success = false;
   assert(!m_isInit);
-  assert(deviceType != PIM_DEVICE_NONE);
-  configDevice(deviceType);
 
   if (!configFileName) {
     std::printf("PIM-Error: Null PIM device config file name\n");
@@ -221,8 +223,12 @@ pimDevice::init(PimDeviceEnum deviceType, const char* configFileName)
     return false;
   }
 
-  m_deviceType = deviceType;
-  
+  // Assign simulation target based on the input argument
+  m_simTarget = deviceType;
+  if (m_simTarget == PIM_FUNCTIONAL) {
+    m_simTarget = PIM_DEVICE_BITSIMD_V;
+  }
+
   // Read file content
   unsigned numRanks;
   unsigned numBankPerRank;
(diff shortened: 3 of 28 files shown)


configs/PIMeval_BitSimdV.cfg
```diff
@@ -0,0 +1,8 @@
+num_ranks = 2
+num_bank_per_rank = 128
+num_subarray_per_bank = 32
+num_row_per_subarray = 1024
+num_col_per_subarray = 8192
+simulation_target = PIM_DEVICE_BITSIMD_V
+max_num_threads = 10
```
fasiddique, line 7: Let's remove this as it will be hardware dependent.

fasiddique (commented):
@mhgholamrezaei thank you for doing this! Could you please add a test example so that I can test it?

configs/PIMeval_BitSimdV.cfg
deyuan: This is upper limit. The actual number of threads is also determined by hardware concurrency. 
Could we support comments in this config file, such as "#..."?

libpimeval/src/libpimeval.h
```diff
@@ -30,6 +30,7 @@ enum PimDeviceEnum {
   PIM_DEVICE_BITSIMD_H,
   PIM_DEVICE_FULCRUM,
   PIM_DEVICE_BANK_LEVEL,
+  PIM_DEVICE_UNKNOWN,
```
deyuan, line 33: We can use PIM_DEVICE_NONE for unknown cases

libpimeval/src/pimSim.cpp
```diff
@@ -55,6 +57,32 @@ pimSim::~pimSim()
...
+  assert(success);
+
+  m_paramsDram = new pimParamsDram(fileContent);
+  m_paramsPerf = new pimParamsPerf(m_paramsDram);
+  m_statsMgr = new pimStatsMgr(m_paramsDram, m_paramsPerf);
+}
+
+//! @brief  Initialize pimSim member classes with default values
+void
+pimSim::init()
```
deyuan, line 79: We need either uninit or handle the case of calling init twice and avoid MLK

libpimeval/src/pimDevice.cpp
```diff
@@ -137,14 +145,27 @@ pimDevice::isHybridLayoutDevice() const
...
-//! @brief  Init pim device, with config file
+//! @brief  Init pim device, with input arguments 
 bool
 pimDevice::init(PimDeviceEnum deviceType, unsigned numRanks, unsigned numBankPerRank, unsigned numSubarrayPerBank, unsigned numRows, unsigned numCols)
 {
   assert(!m_isInit);
   assert(deviceType != PIM_DEVICE_NONE);
 
-  configDevice(deviceType);
+  std::printf("PIM-Info: Trying to read it from envirnment variable %s\n", pimUtils::envVarPimEvalTarget);
```
deyuan, line 155: We don't need to always show this if the env settings are optional

libpimeval/src/pimDevice.cpp
```diff
@@ -137,14 +145,27 @@ pimDevice::isHybridLayoutDevice() const
...
+  // Read envirnment variable for the simulation target
+  std::string pimEvalTarget;
+  if (!pimUtils::getEnvVar(pimUtils::envVarPimEvalTarget, pimEvalTarget)) {
+    std::printf("PIM-Info: Could not read environment variable %s\n", pimUtils::envVarPimEvalTarget);
+  }
+  m_simTarget = pimUtils::strToPimDeviceEnum(pimEvalTarget);
+  if (m_simTarget == PIM_DEVICE_UNKNOWN) {
+    std::printf("PIM-Error: Invalid simulation target %s\n", pimEvalTarget.c_str());
```
deyuan, line 163: Do we still support make PIM_SIM_TARGET?

deyuan (commented):
Added some comments

fasiddique:
I would like to test it before merging. Please let me know once it is ready to test and what are the steps. Until then, please hold off merging.

configs/PIMeval_BitSimdV.cfg
fasiddique: Does this mean that if hardware thread is 8 it will run using 8 cores and if the hardware thread is 32 it will use 10 cores?

fasiddique: Just tested the code and if the hardware has 12 hardware core it is using 10 cores. This is not correct behavior.

fasiddique:
Thank you Hosein! This is really helpful addition to the simulator. I have tested the code and it works! However I have some concerns re behavior if a config file is not given, why are we showing this info and error? Is it intended? I think this is confusing and if no config file is given we should run it using the default and not show the warning.
```
PIM-Info: Trying to read it from envirnment variable PIMEVAL_TARGET
PIM-Info: Could not read environment variable PIMEVAL_TARGET
PIM-Error: Invalid simulation target 
```

configs/PIMeval_BitSimdV.cfg
mhgholamrezaei: @fasiddique Thanks for catching this issue. The right logic should be the following function:
num_threads = min(<user input>, std::thread::hardware_concurrency());
Please, correct me if I am wrong.

mhgholamrezaei: @deyuan, We can add "# ..." as line comment, however DRAMsim3 uses ";" for commenting. My suggestion is to support "#" for line comment and ";" as DRAMsim3 config files. What is you opinion?
[DRAMSim3 config comment example](https://github.com/umd-memsys/DRAMsim3/blob/29817593b3389f1337235d63cac515024ab8fd6e/configs/GDDR5_1Gb_x32.ini#L22C12-L22C50)

libpimeval/src/libpimeval.h
mhgholamrezaei: @deyuan Thank you for identifying this issue. Will be fixed.

libpimeval/src/pimSim.cpp
mhgholamrezaei: @deyuan
Thanks for catching this, MLK. To resolve this issue, I propose adding a new private member field to the pimSim class to track whether init has been called. If init has already been called, an error status should be returned. Do you have any other suggestions?

configs/PIMeval_BitSimdV.cfg
fasiddique: > @fasiddique Thanks for catching this issue. The right logic should be the following function:
> 
> ```
> num_threads = min(<user input>, std::thread::hardware_concurrency());
> ```
> 
> Please, correct me if I am wrong.

Yes.
But what is the purpose of this parameter? Why aren't we directly taking the hardware thread count? Is this to sweep cores for simulator runtime?

libpimeval/src/pimDevice.cpp
mhgholamrezaei: @deyuan Thanks for noticing this bug. I solved it in new commit.

mhgholamrezaei:
@deyuan @fasiddique 
Thanks for the code review.

In the last commit of this PR, the issues with simulation target PIM_DEVICE_UNKNOWN are solved. I need to discuss/clarify the following issues:
- How to resolve the value of thread_max?
- How to handle the MLK caused by `pimDevice::init()`?
- Comment syntax in the config files

fasiddique:
Thanks Hosein for doing this! I pushed a minor change in the util to remove the warning and tested the code afterwards. Please review it. Everything else looks good to me! I am fine with merging this.

mhgholamrezaei:
@fasiddique 
Thanks for fixing this issue.

fasiddique (approved):

libpimeval/src/pimDevice.cpp
```diff
@@ -228,13 +282,81 @@ pimDevice::init(PimDeviceEnum deviceType, const char* configFileName)
   m_resMgr = new pimResMgr(this);
 
   m_cores.resize(m_numCores, pimCore(m_numRows, m_numCols));
+  std::printf("PIM-Info: Current Device = %s, Simulation Target = %s\n",
+              pimUtils::pimDeviceEnumToStr(m_deviceType).c_str(),
```
deyuan, line 286: Where do you assign the m_deviceType variable?

libpimeval/src/pimSim.cpp
```diff
@@ -82,7 +123,9 @@ bool
 pimSim::createDevice(PimDeviceEnum deviceType, unsigned numRanks, unsigned numBankPerRank, unsigned numSubarrayPerBank, unsigned numRows, unsigned numCols)
 {
   pimPerfMon perfMon("createDevice");
+  init();
```
deyuan, line 126: We need to move down this init call after the check below. If there is already a device, we should just return false without calling init or uninit.

libpimeval/src/pimSim.h
```diff
@@ -142,7 +146,12 @@ class pimSim
   pimParamsPerf* m_paramsPerf = nullptr;
   pimStatsMgr* m_statsMgr = nullptr;
   pimUtils::threadPool* m_threadPool = nullptr;
-  unsigned m_numThreads = 1;
+  unsigned m_numThreads = 0;
+  std::string m_memConfigFileName;
+  std::string m_configFilesPath;
+  bool m_initCalled = false;
+
+  static constexpr const char* m_defaultMemConfig = "DDR4_8Gb_x16_3200.ini";
```
deyuan, line 154: Can this be null by default, so that we use the default numbers in pimParamsDram ctor?

libpimeval/src/pimSim.cpp
```diff
@@ -102,18 +148,58 @@ bool
...
+
+  std::string fileContent;
+  success = pimUtils::readFileContent(correctConfigFileName.c_str(), fileContent);
+  assert(success);
+
+  m_configFilesPath = pimUtils::getDirectoryPath(correctConfigFileName);
+
+  init(fileContent);
   if (m_device) {
+    uninit();
```
deyuan, line 188: We cannot call uninit here

deyuan (changes requested):
Looks like there are a few issues need to be validated and resolved

libpimeval/src/pimSim.cpp
mhgholamrezaei: Thanks. Updated.

libpimeval/src/pimSim.h
mhgholamrezaei: Yes, It is updated.

libpimeval/src/pimDevice.cpp
mhgholamrezaei: It is still unclear for me how to resolve the value of m_deviceType based on m_simTarget. Could you please describe how these two variables are mapped together?

libpimeval/src/pimSim.cpp
mhgholamrezaei: Thanks. Fixed!

mhgholamrezaei:
@deyuan 
Currently, all the issues you mentioned have been resolved except for the one related to `m_deviceType`. Could you please review my previous message?

libpimeval/src/pimDevice.cpp
deyuan: Given a simTarget, we need to know if the application is running functional simulation or micro-ops level simulation. In functional simulation, the application can only use high-level functional APIs. In micro-ops level simulation, the application can only use architecture-specific micro-ops. We'll add device type sanity checks to limit the usage later. An application can create a device use PIM_FUNCTIONAL to run functional simulation and a specific device type to run micro-ops level simulation. For now, we can make sure the config file can only affect applications created with PIM_FUNCTIONAL for now.

fasiddique:
@deyuan could you please check this out?

libpimeval/src/pimUtils.h
```diff
@@ -15,18 +15,65 @@
...
+      {PIM_DEVICE_BITSIMD_V_AP, "PIM_DEVICE_BITSIMD_V_AP"},
+      {PIM_DEVICE_DRISA_NOR, "PIM_DEVICE_DRISA_NOR"},
+      {PIM_DEVICE_DRISA_MIXED, "PIM_DEVICE_DRISA_MIXED"},
+      {PIM_DEVICE_SIMDRAM, "PIM_DEVICE_SIMDRAM"},
+      {PIM_DEVICE_BITSIMD_H, "PIM_DEVICE_BITSIMD_H"},
+      {PIM_DEVICE_FULCRUM, "PIM_DEVICE_FULCRUM"},
+      {PIM_DEVICE_BANK_LEVEL, "PIM_DEVICE_BANK_LEVEL"}
+  };
+
+  const std::unordered_map<std::string, PimDeviceEnum> strToEnumMap = {
```
deyuan, line 59: It's better to keep just one unordered_map from enum to string. It's okay to be O(n) when converting from string to enum, since it is only used for config parsing.

libpimeval/src/pimUtils.h
```diff
@@ -15,18 +15,65 @@
...
+      {"PIM_DEVICE_BITSIMD_V_AP", PIM_DEVICE_BITSIMD_V_AP},
+      {"PIM_DEVICE_DRISA_NOR", PIM_DEVICE_DRISA_NOR},
+      {"PIM_DEVICE_DRISA_MIXED", PIM_DEVICE_DRISA_MIXED},
+      {"PIM_DEVICE_SIMDRAM", PIM_DEVICE_SIMDRAM},
+      {"PIM_DEVICE_BITSIMD_H", PIM_DEVICE_BITSIMD_H},
+      {"PIM_DEVICE_FULCRUM", PIM_DEVICE_FULCRUM},
+      {"PIM_DEVICE_BANK_LEVEL", PIM_DEVICE_BANK_LEVEL}
+  };
+
+  static constexpr const char* envVarPimEvalTarget = "PIMEVAL_TARGET";
```
deyuan, line 74: Is this being used?

libpimeval/src/pimSim.h
```diff
@@ -142,7 +146,12 @@ class pimSim
   pimParamsPerf* m_paramsPerf = nullptr;
   pimStatsMgr* m_statsMgr = nullptr;
   pimUtils::threadPool* m_threadPool = nullptr;
-  unsigned m_numThreads = 1;
+  unsigned m_numThreads = 0;
+  std::string m_memConfigFileName;
+  std::string m_configFilesPath;
+  bool m_initCalled = false;
+
+  static constexpr const char* m_defaultMemConfig = "";
```
deyuan, line 154: This variable is not needed

libpimeval/src/pimUtils.h
fasiddique: I actually think @mhgholamrezaei's approach is cleaner and we should keep it.

libpimeval/src/pimSim.cpp
```diff
@@ -41,13 +46,54 @@ pimSim::destroy()
...
+    assert(success);
+
+    if (m_memConfigFileName != m_defaultMemConfig) {
+      std::string memConfigFileFullPath = m_configFilesPath + m_memConfigFileName;
+      std::string fileContent;
+      success = pimUtils::readFileContent(memConfigFileFullPath.c_str(), fileContent);
+      assert(success);
+      m_paramsDram = new pimParamsDram(fileContent);
+    }
+    else {
```
deyuan, line 72: Minor coding style: Suggest putting else inline for consistency and compactness, e.g. "} else {"

libpimeval/src/pimSim.h
```diff
@@ -133,6 +133,10 @@ class pimSim
   ~pimSim();
   pimSim(const pimSim&) = delete;
   pimSim operator=(const pimSim&) = delete;
+  void init();
```
deyuan, line 136: Can we just keep one init function? The simConfig parameter can be empty string by default and act exactly like this

libpimeval/src/pimSim.cpp
```diff
@@ -102,18 +151,57 @@ bool
 pimSim::createDeviceFromConfig(PimDeviceEnum deviceType, const char* configFileName)
 {
   pimPerfMon perfMon("createDeviceFromConfig");
+  bool success = false;
+  std::string correctConfigFileName;
+  if (!configFileName) {
+    std::printf("PIM-Info: Null PIM device config file name. Read the config file name from environment variables %s and %s\n", pimUtils::envVarPimEvalConfigPath, pimUtils::envVarPimEvalConfigSim);
+
+    // Read envirnment variable for the config file path
```
deyuan, line 159: Typo environment

libpimeval/src/pimSim.h
```diff
@@ -133,6 +133,10 @@ class pimSim
...
+  void init(const std::string& simConfig);
```
deyuan, line 137: The parameter name is confusion and does not reflect it needs to be file contents. Can it just be file name string?

libpimeval/src/pimDevice.cpp
```diff
@@ -137,14 +140,36 @@ pimDevice::isHybridLayoutDevice() const
...
+    if (m_simTarget == PIM_DEVICE_NONE) {
+      std::printf("PIM-Warning: Invalid value %s for environment varialbe %s\n", pimEvalTarget.c_str(), pimUtils::envVarPimEvalTarget);
+      readSimTargetFromMakeArgument = true;
+    }
+    if (readSimTargetFromMakeArgument){
+      configSimTarget(m_simTarget);
+    }
+  }
   std::printf("PIM-Info: Current Device = %s, Simulation Target = %s\n",
               pimUtils::pimDeviceEnumToStr(m_deviceType).c_str(),
```
deyuan, line 174: Please assign value to m_deviceType, otherwise it's always NONE here

deyuan (commented):
Thanks for the code changes. Request some further improvements with another round of review.

libpimeval/src/pimSim.h
deyuan: You can keep this or change the other init file to take file name as param

libpimeval/src/pimUtils.h
deyuan: Okay, this is optional.

libpimeval/src/pimUtils.h
mhgholamrezaei: Yes. It is used in `pimDevice.cpp`.

libpimeval/src/pimSim.h
mhgholamrezaei: Thanks. removed!

libpimeval/src/pimSim.cpp
mhgholamrezaei: Thanks, Fixed these issues.

libpimeval/src/pimSim.cpp
mhgholamrezaei: Thanks for catching this. Fixed.

libpimeval/src/pimSim.h
mhgholamrezaei: Thanks. Renamed this variable to `simConfigFileConetnt`.

libpimeval/src/pimSim.h
mhgholamrezaei: Fixed!

mhgholamrezaei:
@deyuan 
Thanks for the code review. Please take a look at the new changes.

deyuan (approved):
Looks great to me, thanks!

PR #64, opened 2024-07-25
restructure knn

kmd2zjw:


PIMbench/cpp-knn/baselines/CPU/knn.cpp
```diff
@@ -0,0 +1,280 @@
...
+ */
+int main(int argc, char **argv)
+{
+    struct Params params = input_params(argc, argv);
+
+
+    if (params.inputTestFile == nullptr)
+    {
+        //getMatrix(params.numTestPoints, params.dimension, 0, testPoints);
+        getMatrix(params.numTestPoints, params.dimension, testPoints);
```
fasiddique, line 243:
```suggestion
        getMatrix<int>(params.numTestPoints, params.dimension, testPoints);
```
Please note that, these functions are declared as templates, so you need to add type whenever you call these functions.

PIMbench/cpp-knn/baselines/CPU/knn.cpp
```diff
@@ -0,0 +1,280 @@
...
+        vector<vector<int>> test_data_int = readCSV(params.inputTestFile);
+        params.dimension = test_data_int[0].size();
+        params.numTestPoints = test_data_int.size();
+
+        testPoints = vector<vector<int>>(test_data_int.begin(), test_data_int.end());;
+    }
+    if (params.inputDataFile == nullptr)
+    {
+        //getMatrix(params.numDataPoints, params.dimension, 0, dataPoints);
+        getMatrix(params.numDataPoints, params.dimension, dataPoints);
```
fasiddique, line 256: Add type to the call

PIMbench/cpp-knn/baselines/CPU/knn.cpp
```diff
@@ -0,0 +1,280 @@
...
+#include <cfloat>
+
+vector<vector<int>> dataPoints;
+vector<vector<int>> testPoints;
+
+// Params ---------------------------------------------------------------------
+typedef struct Params
+{
+    int numTestPoints;
+    int numDataPoints;
```
fasiddique, line 31:
```suggestion
    uint64_t numTestPoints;
    uint64_t numDataPoints;
```
this is to avoid overflow in case of very large number. Could you please update all the usage of this accordingly?

PIMbench/cpp-knn/baselines/CPU/knn.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+    return p;
+}
+
+inline int calculateDistance(const vector<int> &pointA, const vector<int> &pointB, int dim, int target) {
+    int sum = 0;
+    for (int i = 0; i < dim; i++)
+    {
+        if (i == target) {continue;}
+        sum += abs(pointA[i] - pointB[i]);
```
fasiddique, line 117:
```suggestion
        if (i != target)
           sum += abs(pointA[i] - pointB[i]);
```
Please verify this before updating

PIMbench/cpp-knn/baselines/GPU/knncuda.cu
```diff
@@ -0,0 +1,438 @@
...
+
+    // Thread index
+    int tx = threadIdx.x;
+    int ty = threadIdx.y;
+
+    // Initializarion of the SSD for the current thread
+    float ssd = 0.f;
+
+    // int global_y = blockIdx.y * blockDim.y + ty + offset;
+    // int global_x = blockIdx.x * blockDim.x + tx;
```
fasiddique, line 56:
```suggestion
```

PIMbench/cpp-knn/baselines/GPU/knncuda.cu
```diff
@@ -0,0 +1,438 @@
...
+    }
+    if (err2 != cudaSuccess) {
+        printf("ERROR: Memory allocation error for dist_dev: %s\n", cudaGetErrorString(err2));
+    }
+    if (err3 != cudaSuccess) {
+        printf("ERROR: Memory allocation error for index_dev: %s\n", cudaGetErrorString(err3));
+    }
+    if (err4 != cudaSuccess) {
+        printf("ERROR: Memory allocation error for query_labels_dev: %s\n", cudaGetErrorString(err4));
+    }
```
fasiddique, line 286: if you check this after each call you would not need to declare different variable for this. Also if you exit after an error occur following if condition won't be necessary.

fasiddique (changes requested):

PIMbench/cpp-knn/baselines/CPU/knn.cpp
fasiddique: @kmd2zjw I think you have missed this. Could you please check?

PIMbench/cpp-knn/baselines/GPU/inc/PrintOutput.h
```diff
@@ -0,0 +1,47 @@
...
+extern bool noWarmUp;
+
+
+void printOutput(){
+		//Anay change in the format of the output should also change the format of the parser script
+		/*
+        printf(
+            "GBInputPerSecond = %.10f \n ",
+            gigaProcessedInSec);
+            */
```
fasiddique, line 27:
```suggestion
```

PIMbench/cpp-knn/baselines/GPU/knncuda.cu
```diff
@@ -0,0 +1,418 @@
...
+                     int           dim,
+                     int           k,
+                     int *         query_labels, double &elapsed_time) {
+
+    // Constants
+    const unsigned int size_of_float = sizeof(float);
+    const unsigned int size_of_int   = sizeof(int);
+
+    // Return variables
+    cudaError_t err0, err1, err2, err3, err4;
```
fasiddique, line 241: here, one error variable should be fine. As we can reuse it.

PIMbench/cpp-knn/baselines/GPU/Makefile
```diff
@@ -0,0 +1,11 @@
+# Compiler
+CXX := nvcc
+
+# Executable name
+EXEC := knn.out
+
+all:
+	$(CXX) -o $(EXEC) -I./inc -I/sw/ubuntu-22.04/cuda/extras/CUPTI/include -L/sw/ubuntu-22.04/cuda/extras/CUPTI/lib64 -DONLY_TIME_KERNELS test.cpp -I./inc knncuda.cu -lcuda -lcublas -lcudart -Wno-deprecated-gpu-targets
```
fasiddique, line 8: Can you change it to be the same as vector add cuda makefile

kmd2zjw: Somehow this made my execution slower when I made the change but maybe it was something with portal

PIMbench/cpp-knn/baselines/CPU/knn.cpp
```diff
@@ -0,0 +1,268 @@
...
+inline int calculateDistance(const vector<int> &pointA, const vector<int> &pointB, int dim, int target) {
+    int sum = 0;
+    for (int i = 0; i < dim; i++)
+    {
+        if (i != target) {
+            sum += abs(pointA[i] - pointB[i]);
+        }
+    }
+    return sum;
+}
```
fasiddique, line 121:
```suggestion
inline int calculateDistance(const vector<int> &pointA, const vector<int> &pointB, int dim) {
    int sum = 0;
    for (int i = 0; i < dim; i++)
    {
            sum += abs(pointA[i] - pointB[i]);
    }
    return sum;
}
```

kmd2zjw:
Just pushed a fix

kmd2zjw:
I fixed the treading. There was no performance change for some reason but at least its more correct.

fasiddique (approved):

PR #63, opened 2024-07-25
Refactored directories for Histogram and Brightness

hugoabbot:
all tested and working

PR #66, opened 2024-07-26
Clean up warnings in benchmarks

deyuan:
- Do not warn for unused variables for perf build
- Use const char * for file names
- Use %llu to print unsigned long long
- C++ style string comparison


fasiddique:
Let's hold this off until PR #63 gets merged as this is not a critical PR and PR #63 will have conflicts if this gets pushed.

fasiddique (approved):

Issue #16, opened 2024-07-23
Support H-layout padding rules during allocation


deyuan:
H-layout padding rules:
- The pimAlloc API will always allocate elements based on data type, but the actual num bits can be smaller.
- The pimAllocAssociated API will pad elements based on the data width of the reference object. It will error out if new object has larger data width than the reference.

PR #72, opened 2024-07-28
Support get ith element location in a region for mixed data types.

deyuan:
2nd batch of code for supporting mixed data types and element padding:
- Support querying ith element location directly from a region instead of in pimCmd
- Fixed a bug in pimRotateElementsLeft. No impact on benchmarks since it is not used.

No impact on existing behavior. Can be merged before other PRs. 
Related to Issue #16.

### DIFF
--- a/libpimeval/src/pimCmd.cpp
+++ b/libpimeval/src/pimCmd.cpp
@@ -500,8 +500,8 @@ pimCmdFunc1::computeRegion(unsigned index)
   unsigned numElementsInRegion = srcRegion.getNumElemInRegion();
   for (unsigned j = 0; j < numElementsInRegion; ++j) {
     if (dataType == PIM_INT8 || dataType == PIM_INT16 || dataType == PIM_INT32 || dataType == PIM_INT64 || dataType == PIM_UINT8 || dataType == PIM_UINT16 || dataType == PIM_UINT32 || dataType == PIM_UINT64) {
-      auto locSrc = locateNthElement(srcRegion, isVLayout, j, bitsPerElementSrc);
-      auto locDest = locateNthElement(destRegion, isVLayout, j, bitsPerElementDest);
+      auto locSrc = srcRegion.locateIthElemInRegion(j);
+      auto locDest = destRegion.locateIthElemInRegion(j);
       auto operandBits = getBits(core, isVLayout, locSrc.first, locSrc.second, bitsPerElementSrc);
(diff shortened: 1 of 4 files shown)


libpimeval/src/pimCmd.cpp
```diff
@@ -500,8 +500,8 @@ pimCmdFunc1::computeRegion(unsigned index)
   unsigned numElementsInRegion = srcRegion.getNumElemInRegion();
   for (unsigned j = 0; j < numElementsInRegion; ++j) {
     if (dataType == PIM_INT8 || dataType == PIM_INT16 || dataType == PIM_INT32 || dataType == PIM_INT64 || dataType == PIM_UINT8 || dataType == PIM_UINT16 || dataType == PIM_UINT32 || dataType == PIM_UINT64) {
-      auto locSrc = locateNthElement(srcRegion, isVLayout, j, bitsPerElementSrc);
-      auto locDest = locateNthElement(destRegion, isVLayout, j, bitsPerElementDest);
+      auto locSrc = srcRegion.locateIthElemInRegion(j);
+      auto locDest = destRegion.locateIthElemInRegion(j);
```
fasiddique, line 504: how are you deciding if it's vertical or horizontal? as it seems like previously we were passing it on a parameter.

libpimeval/src/pimResMgr.h
```diff
@@ -44,6 +45,14 @@ class pimRegion
   uint64_t getElemIdxBegin() const { return m_elemIdxBegin; }
   uint64_t getElemIdxEnd() const { return m_elemIdxEnd; }
   uint64_t getNumElemInRegion() const { return m_elemIdxEnd - m_elemIdxBegin; }
+  unsigned getNumColsPerElem() const { return m_numColsPerElem; }
+
+  std::pair<unsigned, unsigned> locateIthElemInRegion(unsigned i) const {
+    assert(i < getNumElemInRegion());
+    unsigned rowIdx = m_rowIdx; // only one row of elements per region
+    unsigned colIdx = m_colIdx + i * m_numColsPerElem;
```
fasiddique, line 53: for vertical layout isn't it multiple rows?

libpimeval/src/pimCmd.cpp
deyuan: The key is numColsPerElement. For vertical layout it's 1

libpimeval/src/pimResMgr.h
deyuan: Each region contains one row of elements, no matter if it's V or H layout. You can think each element takes a rectangle area of bits, and here we need to get the location of upper left corner.

fasiddique (approved):

PR #83, opened 2024-07-31
AXPY, GEMM, GEMV refactored directories

hugoabbot:
Added GPU and CPU code for all 3. Everything as been tested and is working on the portal.


fasiddique (approved):
Thank you!

PR #89, opened 2024-08-01
Perf energy model of reduction sum for bit-serial PIM, and various fixes

deyuan:
Code changes:
- Rename pimParamsPerf.h/cc to pimPerfEnergyModels.h/cc
- Fix compilation warnings and code format issues in pimSim/pimDevice
- Fix incorrect uninit during device deletion
- Fix compilation issue of AES app in Makefile
- Update perf and energy model of reduction sum of bit-serial PIM


fasiddique (approved):
Thanks a lot @deyuan!!!

PR #107, opened 2024-08-05
Batched gemv

fasiddique:


misc-bench/cpp-batched-gemv/Makefile
```diff
@@ -0,0 +1,16 @@
+# Makefile: C++ version of matrix matrix multiplication
```
KumareshN93, line 1: I believe the header should be batch multiplication of matrix and vector. Might need to be updated for other files as well.

misc-bench/cpp-batched-gemv/PIM/gemv-batched.cpp
```diff
@@ -0,0 +1,340 @@
+// Test: C++ version of matrix matrix multiplication
```
KumareshN93, line 1: gemm is used in some places in this file..

KumareshN93 (approved):
The PIM code works fine locally at my end.

PR #111, opened 2024-08-06
Old GPU implementation based on Marzieh's repo

kmd2zjw:


kmd2zjw:
See newer PR for up to date version.

PR #122, opened 2024-08-10
Bugfix energy

fasiddique:


fasiddique:
@alifahmed @fzx333 Please check this and let me know

libpimeval/src/pimPerfEnergyModels.cpp
```diff
@@ -94,7 +95,7 @@ pimParamsPerf::getPerfEnergyBitSerial(PimDeviceEnum deviceType, PimCmdEnum cmdTy
         if (it3 != it2->second.end()) {
           const auto& [numR, numW, numL] = it3->second;
           msRuntime += m_tR * numR + m_tW * numW + m_tL * numL;
-          mjEnergy += (m_eL * numL * obj.getMaxElementsPerRegion() * obj.getBitsPerElement()) * obj.getNumCoresUsed() + (m_eAP * numR + m_eAP * numW) * m_numChipsPerRank * numRanks;
+          mjEnergy += ((m_eL * numL * obj.getMaxElementsPerRegion()) + (m_eAP * numR + m_eAP * numW)) * numCores;
```
fzx333, line 98: Just to confirm that "(m_eL * numL * obj.getMaxElementsPerRegion()" should also be multiplied with numCore, right?

fasiddique: Yes. Fixing this

fasiddique: Ok, it's been multiplied.

libpimeval/src/pimPerfEnergyModels.cpp
```diff
@@ -326,8 +324,7 @@ pimParamsPerf::getPerfEnergyForFunc2(PimCmdEnum cmdType, const pimObjInfo& obj)
        * As a result, only one read operation and one write operation is necessary for the entire pass.
       */
       msRuntime = m_tR + m_tW + (maxElementsPerRegion * numberOfALUOperationPerElement * m_fulcrumAluLatency * 2) * numPass;
-      mjEnergy = obj.getNumCoresUsed() * numPass * ((maxElementsPerRegion - 1) * 3 *  m_fulcrumShiftEnergy) + ((maxElementsPerRegion) * m_fulcrumALUArithmeticEnergy * numberOfALUOperationPerElement);
-      mjEnergy += m_eAP * 3 * m_numChipsPerRank * numRanks;
+      mjEnergy = numCoresUsed * numPass * ((m_eAP * 3) + ((maxElementsPerRegion - 1) * 3 *  m_fulcrumShiftEnergy) + ((maxElementsPerRegion) * m_fulcrumALUArithmeticEnergy * numberOfALUOperationPerElement));
       mjEnergy += m_pBChip * m_numChipsPerRank * numRanks * msRuntime;
```
fzx333, line 328: The left side code is the change I made last time, and the reason I did that is because I was trying to avoid multiplying m_eAP with numCore, and only multiply it by the m_numChipsPerRank * numRanks, now it seems like I should have wrote mjEnergy += numPass * m_eAP * 3 * m_numChipsPerRank * numRanks; in line 330.

But since you are already getting a reasonable number for fulcrum, I think it would be better to just keep the changes you did! (which is a very worst case senario)

libpimeval/src/pimPerfEnergyModels.cpp
```diff
@@ -371,8 +368,7 @@ pimParamsPerf::getPerfEnergyForFunc2(PimCmdEnum cmdType, const pimObjInfo& obj)
       double totalGDLOverhead = m_tGDL * numGDLItr * 2; // one read can be pipelined
       msRuntime = 2 * m_tR + m_tW + totalGDLOverhead + maxElementsPerRegion * m_blimpCoreLatency * numberOfOperationPerElement;
       msRuntime *= numPass;
-      mjEnergy = m_eAP * 3 * m_numChipsPerRank * numRanks + (m_eGDL * numGDLItr * 3 + (maxElementsPerRegion * m_blimpArithmeticEnergy * numberOfOperationPerElement)) * obj.getNumCoresUsed();
-      mjEnergy *= numPass;
+      mjEnergy = ((m_eAP * 3) + (m_eGDL * numGDLItr * 3 + (maxElementsPerRegion * m_blimpArithmeticEnergy * numberOfOperationPerElement))) * numCoresUsed * numPass;
```
fzx333, line 371: I remember that m_eGDL is the energy for bringing an entire row to the bank-level SA, right? If that's the case, it shouldn't be multiplied with numCores, right? that would make it so big.

fzx333, line 371: If this is bank-level energy, I actually feel like nothing here should be multiplied with numCores, right? It should be multiplied by the #ALU/bank * #bank

fasiddique: So it shouldn't be multiply by `numGDLItr`?

fasiddique: Shouldn't we multiply by num of banks?

fzx333: I thought the numCores is the the number of subarrays, right? If this is bank-level ALU, then I think it has nothing to do with the number of subarrays.

fzx333: I forgot it, but could it be that numCore represents #banks in the bank-level architecture? Is that the reason you wrote the code in that way? I forgot if you mentioned this, but I just want to make sure we are using numCores correctly here

fasiddique: yes numcores = numbanks for bank-level

alifahmed (approved):
Looks good to me.

libpimeval/src/pimPerfEnergyModels.cpp
```diff
@@ -394,8 +390,8 @@ pimParamsPerf::getPerfEnergyForFunc2(PimCmdEnum cmdType, const pimObjInfo& obj)
       */
       double totalGDLOverhead = m_tGDL * numGDLItr; // both read can be pipelined as multiplication and addition takes twice the time to execute.
       msRuntime = m_tR + (m_tW + totalGDLOverhead + maxElementsPerRegion * m_blimpCoreLatency * numberOfOperationPerElement * 2) * numPass;
-      mjEnergy = m_eAP * 3 * m_numChipsPerRank * numRanks + (m_eGDL * numGDLItr * 3 + (maxElementsPerRegion * m_blimpArithmeticEnergy * numberOfOperationPerElement)) * obj.getNumCoresUsed();
-      mjEnergy += maxElementsPerRegion * numberOfOperationPerElement * m_blimpArithmeticEnergy * obj.getNumCoresUsed();
+      mjEnergy = ((m_eAP * 3) + (m_eGDL * 3 + (maxElementsPerRegion * m_blimpArithmeticEnergy * numberOfOperationPerElement))) * numCoresUsed;
+      mjEnergy += maxElementsPerRegion * numberOfOperationPerElement * m_blimpArithmeticEnergy * numCoresUsed;
```
fzx333, line 394: This part of the code looks good to me now. I wonder if we run it now whether the number makes sense for the bank level. And just to make sure we are doing the same changes to all the bank-level code.

fasiddique:
Thank you so much @alifahmed @fzx333 for helping with this!!!

PR #125, opened 2024-08-19
[misc-bench] refactored Makefiles and file names for continuity

hugoabbot:
Changed Makefiles to use C++17 instead of 11, with some other minor changes to keep the directory consistent.


fasiddique (approved):
Thank you! This looks great.

PR #127, opened 2024-08-20
Added CPU/GPU implementation of convolution and misc. updates in conv.cpp file

KumareshN93:
* CPU/GPU code for convolution prints output like the following example:
>>  python3 conv.py  -b 8
>>  [INFO] Using device: cpu
>>  [INFO] Time taken for convolution for batch size of 8: 100.381851 ms


fasiddique (approved):
Thank you!

PR #128, opened 2024-08-21
Added CPU/GPU implementation of max pooling and misc. updates in pool.cpp file

KumareshN93:
* CPU/GPU code for max pooling prints output like the following example:
  >> python3 pool.py -b 64
  >> [INFO] Using device: cpu
  >> [INFO] Time taken for max pooling for batch size of 64: 348.115206 ms


fasiddique (approved):
Thank you!

PR #129, opened 2024-08-22
Support padding during allocation

deyuan:
Partial implementation:
- Allow bitsPerElement to be smaller than data type bit width
- Allow a narrower data type to be associated to a wider data type

Need to revisit all code depending on bitsPerElement to make it work.

Related to Issue #16


libpimeval/src/pimResMgr.cpp
```diff
@@ -102,6 +102,9 @@ pimResMgr::~pimResMgr()
 }
 
 //! @brief  Alloc a PIM object
+//!         Expect bitsPerElement to be smaller than or equal to the width of dataType
+//!         For V layout, allocate bitsPerElement rows
+//!         For H layout, pad element to the width of dataType
 PimObjId
 pimResMgr::pimAlloc(PimAllocEnum allocType, uint64_t numElements, unsigned bitsPerElement, PimDataType dataType)
```
fasiddique, line 109: Do we actually need `bitsPerElement`? Isn't it redundant? We can derive bits per element from the datatype.

libpimeval/src/pimResMgr.cpp
```diff
@@ -102,6 +102,9 @@ pimResMgr::~pimResMgr()
 }
 
 //! @brief  Alloc a PIM object
+//!         Expect bitsPerElement to be smaller than or equal to the width of dataType
```
fasiddique, line 105: I am wondering why this should be smaller that the datatype width?

libpimeval/src/pimResMgr.cpp
deyuan: This is especially needed for V layout architectures. We support a set of common data types, but an application may just use fewer bits of it. For example, we can allocate an INT32 object but only use 20 bits of it. In this case, this bitsPerElement is essential for performance modeling.

libpimeval/src/pimResMgr.cpp
deyuan: We need to support both V and H layouts. Data types determines the allocated bit width, while bitsPerElements determines how many bits are actually used.

fasiddique: This approach seems quite confusing and increases the risk of introducing bugs. If someone is creating a vector of int32, they should expect it to use 32 bits. If a smaller bit-size is needed, they should opt for a different data type that uses fewer than 32 bits.

libpimeval/src/pimResMgr.cpp
fasiddique: Then we should introduce INT20. I don't think this is a good design at all.

deyuan: I think we've discussed the following in emails. We need to support pimAllocAssociate with different data types, and PIM operations with mixed data types. For example, we may allocate an INT32 object first, then associate an INT8 or BOOL object with it. In this case, we would like to do element padding and align elements with different types. Each element will have a data width and a padded width. The code changes here are for supporting such functionality. We can remove bitsPerElements from the functional API later for simplicity, but I don't think code design is bad here

libpimeval/src/pimResMgr.cpp
deyuan: Let's discuss how to support element padding when you are back

deyuan:
Switch development to dev-alloc branch

PR #133, opened 2024-08-22
Added CPU/GPU implementation of ReLU and misc. updates in relu.cpp file

KumareshN93:
* CPU/GPU code for ReLU prints output like the following example:
  >> python3 relu.py -b 64
  >> [INFO] Using device: cpu
  >> [INFO] Time taken for ReLU for batch size of 64: 177.109957 ms

* Updated utilML.h file per the changes in the kernels' cpp files.


PIMbench/cpp-relu/PIM/relu.cpp
```diff
@@ -259,72 +254,74 @@ int main(int argc, char *argv[]) {
...
+  if (status != PIM_OK) {
+    std::cout << "Abort: pimGetDeviceProperties failed" << std::endl;
+    return 1;
+  }
+  // Get the device parameters
+  uint64_t numCols = deviceProp.numColPerSubarray;
+  uint64_t numRows = deviceProp.numRowPerSubarray;
+  uint64_t numOfBits = uint64_t(deviceProp.numRanks) * uint64_t(deviceProp.numBankPerRank) * uint64_t(deviceProp.numSubarrayPerBank) * numCols * numRows;
+
+  uint64_t numOfPIMRow = 1;
```
fasiddique, line 268: Why is this value 1?

PIMbench/cpp-relu/PIM/relu.cpp
```diff
@@ -259,72 +254,74 @@ int main(int argc, char *argv[]) {
...
+  uint64_t numOfMatPerRow = std::min(static_cast<uint64_t>(std::floor((1.0 * numOfBits) / (inputHeight * inputWidth * 32))), static_cast<uint64_t>(inputWidth));
```
fasiddique, line 269: Is `32` here the size of data type?

PIMbench/cpp-relu/PIM/relu.cpp
```diff
@@ -259,72 +254,74 @@ int main(int argc, char *argv[]) {
...
   // Loop through input depth in chunks
-  for (int j = 0; j < inputDepth; j += numOfMatPerRow) {
-    int matChunk = (numOfMatPerRow + j) <= inputDepth ? (numOfMatPerRow + j) : inputDepth;
+  for (uint64_t j = 0; j < inputDepth; j += numOfMatPerRow) {
+    uint64_t matChunk = (numOfMatPerRow + j) <= inputDepth ? (numOfMatPerRow + j) : inputDepth;
     // Decompose and merge matrices
-    for (int k = j; k < matChunk; k++) {
-      decomposeMatrix(params.row, params.column, 1, 1, 1, 0, inputMatrix[k], decompMat);
+    for (uint64_t k = j; k < matChunk; k++) {
+      decomposeMatrix(inputHeight, inputWidth, 1, 1, 1, 0, inputMatrix[k], decompMat);
```
fasiddique, line 284: Can you please add some comments on why these values are 1, 1, 1, 0

KumareshN93: Added comments and updated the file.

PIMbench/cpp-relu/PIM/relu.cpp
KumareshN93: Yes, 32 is the size of the data type. Instead of hardcoding it, declared it as a variable, and updated the file accordingly.

PIMbench/cpp-relu/PIM/relu.cpp
KumareshN93: '1' here indicates that ReLU operation is an element-wise operation. So, for each element, we do max(0, element) operation.  Hence, the variable is assigned a value of 1.

fasiddique (approved):
Thank you!

PR #135, opened 2024-08-23
Add Batched GeMV Kernel Implementation for CPU, GPU, and PIM

mhgholamrezaei:
**Changes:**

- Added CPU and GPU implementations for the batched GeMV kernel.
    - The CPU and GPU codebases are derived from the existing GeMV CPU and GPU implementations. The only modification made was the addition of an outer loop to support batch processing.
- Added a PIM implementation for the batched GeMV kernel.
- Fixed a bug in the ranged copy function.


libpimeval/src/pimCmd.cpp
fasiddique, line 1: Could you please submit a separate PR for any changes made to the command class? I don't think it is a good idea to add this changes with the batched gemv implementation.

fasiddique (changes requested):
Is the command class changes required for this benchmark implementation? If yes, then I suggest making the changes in command class in a separate PR and adding the gemv implementation in a subsequent PR.

libpimeval/src/pimCmd.cpp
mhgholamrezaei: Thanks for reviewing this PR. I close this one and create two new PRs, one for the bug fix, and one for the batched GeMV kernel.

PR #140, opened 2024-09-04
Add Copy Benchmark

Arleee1:


misc-bench/cpp-copy/baselines/CPU/copy.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include <getopt.h>
+#include <assert.h>
+#include <stdint.h>
+#include <iomanip>
+#include <chrono>
+#include <omp.h>
+#include <iostream>
+#include <vector>
+
+#include "utilBaselines.h"
```
fasiddique, line 19: The util file is in the misc-bench directory. So I assume this include will not work?

Arleee1: Yes, as I have included the directory in the include path in the Makefile. I can reformat it to be like the others if you would like though.

misc-bench/cpp-copy/baselines/GPU/copy.cu
```diff
@@ -0,0 +1,171 @@
...
+    gpu_y[idx] = gpu_x[idx];
+  }
+}
+
+/**
+ * @brief Main of the Host Application.
+ */
+int main(int argc, char **argv)
+{
+  cudaGetLastError();
```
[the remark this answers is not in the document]
Arleee1: I have removed it, I apologize.

misc-bench/cpp-copy/baselines/CPU/copy.cpp
fasiddique: This is fine, thank you!

misc-bench/cpp-copy/baselines/GPU/copy.cu
```diff
@@ -0,0 +1,169 @@
...
+int main(int argc, char **argv)
+{
+  struct Params params = input_params(argc, argv);
+
+  uint64_t vectorSize = params.vectorLength;
+
+  std::vector<int> X;
+  std::vector<int> Y;
+
+  getVector(vectorSize, X);
```
fasiddique, line 90: If I am not mistaken, `getVector` is a template function so you need to call it as following:

```suggestion
  getVector<int>(vectorSize, X);
```

fasiddique: Could you please fix this in both PRs?

fasiddique (approved):
Thank you!

PR #142, opened 2024-09-05
Fix the Bug in Ranged Data Copy Command

mhgholamrezaei:
Changes:
- Fixed a bug in ranged data copy command

### DIFF
--- a/libpimeval/src/pimCmd.cpp
+++ b/libpimeval/src/pimCmd.cpp
@@ -343,34 +343,38 @@ pimCmdCopy::computeRegion(unsigned index)
     if (m_copyType == PIM_COPY_V) {
       size_t bitIdx = 0;
       for (size_t i = 0; i < (size_t)numAllocRows * numAllocCols; ++i) {
-        bool val = bits[bitIdx++];
-        if (isDCCN) { val = !val; }
         unsigned row = rowIdx + i % numAllocRows;
         unsigned col = colIdx + i / numAllocRows;
         if (!m_copyFullRange) {
           uint64_t currIdx = regionBeginIdx + col;
           if (currIdx >= m_idxBegin && currIdx < m_idxEnd) {
+            bool val = bits[bitIdx++];
+            if (isDCCN) { val = !val; }
             core.setBit(row, col, val);
           }
         } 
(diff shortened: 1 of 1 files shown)


fasiddique (commented):
Thank you! I am wondering if following would be more concise

```
size_t bitIdx = 0;
for (size_t i = 0; i < numAllocRows * numAllocCols; ++i) {
    unsigned row = (m_copyType == PIM_COPY_V) ? (rowIdx + i % numAllocRows) : (rowIdx + i / numAllocCols);
    unsigned col = (m_copyType == PIM_COPY_V) ? (colIdx + i / numAllocRows) : (colIdx + i % numAllocCols);

    if (!m_copyFullRange) {
        uint64_t currIdx = (m_copyType == PIM_COPY_V) ? (regionBeginIdx + col) : (regionBeginIdx + col / bitsPerElement);
        if (currIdx >= m_idxBegin && currIdx < m_idxEnd) {
            bool val = bits[bitIdx++];
            if (isDCCN) { val = !val; }
            core.setBit(row, col, val);
        }
    } else {
        bool val = bits[bitIdx++];
        if (isDCCN) { val = !val; }
        core.setBit(row, col, val);
    }
}
```
There could be some issues with this suggestion
1. I haven't checked the entire file so I am not sure if this will break any other code. This needs to be tested carefully.
2. This suggestion assumed PIM_COPY type will always be either horizontal or vertical. We could also check this at the beginning where bitIdx is set to zero.
3. The following lines can be wrapped in an inline function call to avoid code duplication
```
bool val = bits[bitIdx++];
if (isDCCN) { val = !val; }
core.setBit(row, col, val);
```
@deyuan any thoughts?

deyuan (approved):

deyuan:
> @deyuan any thoughts?
The bug fix looks correct to me. @mhgholamrezaei You can merge this PR to main when ready. It would be great to make sure the test-large-copy-ranged testcase can cover this issue.

I think it's okay to have slightly duplicated code here for now. Merging V and H code is fine as long as it can improve maintainability. We can do separate code cleanup once element padding is fully done, which may affect the way of calculating element location.

PR #150, opened 2024-09-11
Keep one copy of Makefile.common at project root.

deyuan:
PR: dev-benchmark -> main
This branch has reached a stable state that is good for merging to main.

Main changes:
- Keep a single copy of Makefile.common at project root
- Consistent way of using PROJ_ROOT to locate Makefile.common in sub-module Makefiles

No impact on existing behavior.


fasiddique (approved):
LGTM!

PR #156, opened 2024-09-19
Break perf energy models into multiple classes

deyuan:
PR to dev-cleanup branch for early code review and incremental development.

Reorganize code of perf energy modes:
- Rename pimParamsPerf::perfEnergy to pimNS::perfEnergy
- Define pimPerfEnergyFactory
- Define pimPerfEnergyBase, pimPerfEnergyBitSerial, pimPerfEnergyFulcrum, pimPerfEnergyBankLevel
- Move perfEnergyModel from pimSim to pimDevice because of sim target dependency
- Update related code in pimSim, pimDevice, pimCmd, pimStats

Related to Discussion #143 

Warning: It's hard to diff pimPerfEnergyModels.h/cpp. There is no change on perf and energy formulas.

### DIFF
--- a/libpimeval/src/pimPerfEnergyModels.cpp
+++ b/libpimeval/src/pimPerfEnergyModels.cpp
@@ -10,6 +10,38 @@
 #include "pimPerfEnergyTables.h"
 #include <cstdio>
 
+
+//! @brief  A factory function to get perf energy model for sim target
+pimPerfEnergyBase*
+pimPerfEnergyFactory::getPerfEnergyModel(PimDeviceEnum simTarget, pimParamsDram* paramsDram)
+{
+  pimPerfEnergyBase* model = nullptr;
+  switch (simTarget) {
+    case PIM_DEVICE_BITSIMD_V:
+    case PIM_DEVICE_BITSIMD_V_AP:
+    case PIM_DEVICE_BITSIMD_H:
+    case PIM_DEVICE_SIMDRAM:
+      std::printf("PIM-Info: Created performance energy model for bit-serial PIM\n");
+      model = new pimPerfEnergyBitSerial(paramsDram);
+      break;
+    case PIM_DEVICE_FULCRUM:
+      std::printf("PIM-Info: Created performance energy model for Fulcrum\n");
+      model = new pimPerfEnergyFulcrum(paramsDram);
+      break;
+    case PIM_DEVICE_BANK_LEVEL:
+      std::printf("PIM-Info: Created performance energy model for bank-level PIM\n");
+      model = new pimPerfEnergyBankLevel(paramsDram);
+      break;
+    default:
+      std::printf("PIM-Info: Created performance energy model for base\n");
+      model = new pimPerfEnergyBase(paramsDram);
+  }
+  return model;
+}
+
+///////////////////////////////////////////////////////////////////////////////
+// MODEL: Base
+///////////////////////////////////////////////////////////////////////////////
 //! @brief  pimPerfEnergyBase ctor
 pimPerfEnergyBase::pimPerfEnergyBase(pimParamsDram* paramsDram)
   : m_paramsDram(paramsDram)
@@ -28,11 +60,10 @@ pimPerfEnergyBase::pimPerfEnergyBase(pimParamsDram* paramsDram)
   m_numChipsPerRank = m_paramsDram->getNumChipsPerRank();
 }
 
-//! @brief  Get ms runtime for bytes transferred between host and device
+//! @brief  Perf energy model of data transfer between CPU memory and PIM memory
 pimNS::perfEnergy
 pimPerfEnergyBase::getPerfEnergyForBytesTransfer(PimCmdEnum cmdType, uint64_t numBytes) const
 {
-
   double mjEnergy = 0.0;
   unsigned numRanks = pimSim::get()->getNumRanks();
   double typicalRankBW = m_paramsDram->getTypicalRankBW(); // GB/s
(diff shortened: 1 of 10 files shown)


libpimeval/src/pimPerfEnergyModels.h
```diff
@@ -13,14 +13,7 @@
...
 
-//! @class  pimParamsPerf
-//! @brief  PIM performance parameters base class
-class pimParamsPerf
-{
-public:
-  pimParamsPerf(pimParamsDram* paramsDram);
-  ~pimParamsPerf() {}
-
+namespace pimNS {
```
fasiddique, line 16: What does NS stand for? Also, could you please add some documentation on why this namespace was introduced and what to add under this and what not to add?

libpimeval/src/pimPerfEnergyModels.h
```diff
@@ -63,16 +67,73 @@ class pimParamsPerf
   double m_pBCore; // background power for each core in W
   double m_pBChip; // background power for each core in W
   double m_eGDL = 0.0000102; // CAS energy in mJ
+};
+
+//! @class  pimPerfEnergyBitSerial
+//! @brief  PIM performance energy model for bit-serial PIM family
+class pimPerfEnergyBitSerial : public pimPerfEnergyBase
```
fasiddique, line 74: I really like this class based on PIM architecture. Thank you!

libpimeval/src/pimPerfEnergyModels.h
fasiddique: I assume NS = Name Space? But as I'm reading the code, it seems to evoke the idea of nanoseconds (due to its common use). To avoid confusion, could we consider using something more descriptive like peval or pim?

libpimeval/src/pimPerfEnergyModels.cpp
```diff
@@ -10,8 +10,40 @@
...
+pimPerfEnergyFactory::getPerfEnergyModel(PimDeviceEnum simTarget, pimParamsDram* paramsDram)
+{
+  pimPerfEnergyBase* model = nullptr;
+  switch (simTarget) {
+    case PIM_DEVICE_BITSIMD_V:
+    case PIM_DEVICE_BITSIMD_V_AP:
+    case PIM_DEVICE_BITSIMD_H:
+    case PIM_DEVICE_SIMDRAM:
+      std::printf("PIM-Info: Created performance energy model for bit-serial PIM\n");
+      model = new pimPerfEnergyBitSerial(paramsDram);
```
fasiddique, line 25: I would prefer smart pointers instead of `new` and `delete` to avoid manual memory management.

libpimeval/src/pimPerfEnergyModels.cpp
```diff
@@ -10,8 +10,40 @@
 #include "pimPerfEnergyTables.h"
 #include <cstdio>
 
-//! @brief  pimParamsPerf ctor
-pimParamsPerf::pimParamsPerf(pimParamsDram* paramsDram)
+
+//! @brief  A factory function to get perf energy model for sim target
+pimPerfEnergyBase*
+pimPerfEnergyFactory::getPerfEnergyModel(PimDeviceEnum simTarget, pimParamsDram* paramsDram)
```
fasiddique, line 16:
```suggestion
pimPerfEnergyFactory::getPerfEnergyModel(PimDeviceEnum simTarget, const pimParamsDram& paramsDram)
```
This is to avoid accidental changes to the parameters

libpimeval/src/pimPerfEnergyModels.cpp
```diff
@@ -125,107 +205,260 @@ pimParamsPerf::getPerfEnergyBitSerial(PimDeviceEnum deviceType, PimCmdEnum cmdTy
...
+//! @brief  Perf energy model of Fulcrum for func1
+pimNS::perfEnergy
+pimPerfEnergyFulcrum::getPerfEnergyForFunc1(PimCmdEnum cmdType, const pimObjInfo& obj) const
+{
+  double msRuntime = 0.0;
+  double mjEnergy = 0.0;
+  unsigned numPass = obj.getMaxNumRegionsPerCore();
+  unsigned bitsPerElement = obj.getBitsPerElement();
+  unsigned numCores = obj.getNumCoresUsed();
+  unsigned numRanks = pimSim::get()->getNumRanks();
```
fasiddique, line 432: These are going to be duplicated. I wonder if we should add these in the constructor to avoid duplication?

libpimeval/src/pimPerfEnergyModels.cpp
fasiddique, line 1: This file seems to be very large and it's very difficult to read and track the code. I am wondering if we should follow the following structure?
```
src/
├── models/
│   ├── pimPerfEnergyBase.h
│   ├── pimPerfEnergyBase.cpp
│   ├── pimPerfEnergyBitSerial.h
│   ├── pimPerfEnergyBitSerial.cpp
│   ├── pimPerfEnergyFulcrum.h
│   ├── pimPerfEnergyFulcrum.cpp
│   └── pimPerfEnergyBankLevel.h
│   └── pimPerfEnergyBankLevel.cpppimModelFactory.cpp
```

libpimeval/src/pimPerfEnergyModels.h
deyuan: Renamed to pimeval for clarity

libpimeval/src/pimPerfEnergyModels.cpp
deyuan: Yes. A params class is needed since pimDevice is not fully constructed during perf energy model ctor.

PR #161, opened 2024-09-26
filter by key openmp guard

### LINKED ISSUE #157: Guard openmp code in filter-by-key app

fasiddique:
Closes #157

### DIFF
--- a/PIMbench/filter-by-key/PIM/Makefile
+++ b/PIMbench/filter-by-key/PIM/Makefile
@@ -6,11 +6,14 @@
 PROJ_ROOT = ../../..
 include ${PROJ_ROOT}/Makefile.common
 
+# make USE_OPENMP=1
+USE_OPENMP ?= 0
+ifeq ($(USE_OPENMP),1)
+	CXXFLAGS += -fopenmp
+endif
+
 EXEC := filter.out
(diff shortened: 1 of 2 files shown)


deyuan (changes requested):
This application has below code in db-filtering.cpp which requires special handling:
        int tid = omp_get_thread_num();
@fzx333 mentioned this is no longer needed. We need to clean up both to pass compilation without OpenMP.

fasiddique:
> This application has below code in db-filtering.cpp which requires special handling: int tid = omp_get_thread_num(); @fzx333 mentioned this is no longer needed. We need to clean up both to pass compilation without OpenMP.

@deyuan  I can see that this is used to divide task among the threads. @fzx333, will you please confirm?

fzx333 (approved):

fzx333:
We can just add the openMP guard for now for the purpose of cleanness, and the code will also work. In terms of the changes I mentioned, it might involve more code changes, and I will do them in a future PR.

deyuan (approved):
The changes are good but not complete. Without handling omp_get_thread_num(), the app cannot be compiled without OpenMP library. I'm okay to merge this and let @fzx333 update the code separately.

PR #162, opened 2024-09-30
Fix reinterpret_cast warning

### LINKED ISSUE #6: Fix reinterpret_cast warning issues

Morteza1814:
Fixes #6 
1. Implemented the operandCast function in the pimUtil.h 
2. Used operandCast for casting in the pimCmd.h
3. Used operandCast for asting in pimCmd.cpp for float casting
4. Tested using `./test-functional.out`.


libpimeval/src/pimUtils.h
```diff
@@ -41,6 +42,20 @@ namespace pimUtils
   std::string getDirectoryPath(const std::string& filePath);
   bool getEnvVar(const std::string &varName, std::string &varValue);
 
+  template<typename T> 
+  T operandCast(const void* inputOperand, uint8_t numBytes) {
```
deyuan, line 46: How about we derive numBytes from the typename T here? If this works, we don't need to specify sizeof() in every function call

deyuan: Maybe call it typeCast

deyuan (commented):
Added a few suggestions about operandCast. Could you take a look?

libpimeval/src/pimUtils.h
Morteza1814: Thanks. Updated the code!

deyuan: I feel something is missing here. If the size of inputOperand is less than the size of T, will some garbage value go into the temp variable? Do you think we need to use num bytes of inputOperand but not num bytes of T?

Morteza1814: The inputOperand and T are the same size in all the use cases.

libpimeval/src/pimUtils.h
```diff
@@ -42,6 +43,20 @@ namespace pimUtils
   std::string getDirectoryPath(const std::string& filePath);
   bool getEnvVar(const std::string &varName, std::string &varValue);
 
+  template<typename T> 
+  T typeCast(const void* inputOperand) {
+    T operandValue = 0;  // Ensure operandValue is initialized
+    T temp;              // Temporary variable of type T
+
```
fasiddique, line 50:
```suggestion
assert(inputOperand != nullptr); 
```

libpimeval/src/pimUtils.h
```diff
@@ -42,6 +43,20 @@ namespace pimUtils
...
+    // Copy the specified number of bytes into temp
+    memcpy(&temp, inputOperand, sizeof(T));
+
+    // Assign the value to operandValue (handles sign extension if T is signed)
+    operandValue = static_cast<T>(temp);
```
fasiddique, line 55: do we need the `temp` variable at all?
can we not directly copy to the `operandValue`?

fasiddique (commented):
@Morteza1814 this looks great! Thank you so much for doing this! I have left a couple of minor comments. Would you please check and let me know of your thoughts.

libpimeval/src/pimUtils.h
fasiddique: I am not sure if memcpy will be able to handle the sign extension. If it can't then we can keep it as it is.

libpimeval/src/pimUtils.h
deyuan: As a code designing principle, a utility function needs to be self contained. It's better to avoid creating unnecessary assumptions on inputs. Once PR #172 is merged, we need to run that full functional testing.

libpimeval/src/pimUtils.h
Morteza1814: We need temp for final static_cast which is responsible for sign extension.

fasiddique (approved):

deyuan:
Can we please hold on merging this? I'm looking into a better way to organize the code, and I'll update to you later today.

deyuan:
To honor the contributions of @Morteza1814. Let's merge this. I'll resolve any conflicts in PR #175.

PR #164, opened 2024-10-02
[dev-cleanup] Break perf energy models into multiple classes - Batch 2

deyuan:
PR to dev-cleanup branch for early code review and incremental development.

Changes:
- Further split perf energy models into multiple files
- Rename namespace to pimeval
- Use unique_ptr other than raw pointers
- Define a pimPerfEnergyModelParams to avoid querying partially constructed pimDevice during perf energy model ctors

Related Discussion: #143


fasiddique (approved):
Thank you!!! This looks great!

PR #165, opened 2024-10-02
[dev-alloc] 

deyuan:
PR to dev-alloc for early review and incremental development.

Changes:
- Remove bitsPerElement parameter from pimAlloc and pimAllocAssociated. Auto derive bitsPerElement from data type.
- Add pimAllocCustomized and pimAllocAssociatedCustomized for supporting different bitsPerElement as needed. 
- pimResMgr changes from previous PR #129 
- Add pimUtils::getNumBitsOfDataType
- Add test-padding testcase
- Add misc-bench folder to top-level Makefile
- Fix a but in scale and warnings in bitmap and xor

Related issue: #16

PR #168, opened 2024-10-03
[main] Remove bitsPerElement parameter from default pimAlloc and pimAllocAssociated

deyuan:
PR #165 combines multiple changes to dev-alloc branch, which is hard to review. I would suggest we merge this batch of changes to main first for API updates that touch all benchmark apps. 

List of changes:
- Remove bitsPerElement parameter from pimAlloc and pimAllocAssociated. Auto derive bitsPerElement from data type by calling newly added pimUtils::getNumBitsOfDataType. Currently all bitsPerElement parameters match with given data type exactly, so they can be safely removed.
- Add misc-bench folder to top-level Makefile. Currently it is not compiled.
- Fix a bug in scale (incorrect associated obj) and print warnings in bitmap and xor.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@
 LIBDIR := libpimeval
 BITSERIALDIR := bit-serial
 APPDIR := PIMbench
-TESTDIR := tests
+TESTDIR := misc-bench tests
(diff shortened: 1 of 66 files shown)


fasiddique (approved):
Thank you!

PR #169, opened 2024-10-04
PIMeval: Refactored DRAM Param Class

fasiddique:
### Description

This update refactors the `pimParamDRAM` design to account for differences in timing and energy parameters based on the DRAM protocol. The existing design assumed that all protocols would have the same parameters and that row read/write latency and energy calculation formulas would not vary across protocols. To provide more flexibility, the class structure has been redesigned using the factory pattern. Distinct child classes have been created for each protocol, ensuring the correct parameters are applied. Additionally, a new enum has been introduced to define different protocols.

### Tests

- Executed vector addition, radix sort, and AXPY with default parameters.
- Tested with various configuration files to ensure compatibility.

### DIFF
--- a/libpimeval/src/pimParamsDDRDram.cpp
+++ b/libpimeval/src/pimParamsDDRDram.cpp
@@ -152,3 +152,4 @@ pimParamsDDRDram::pimParamsDDRDram(std::unordered_map<std::string, std::string> 
     throw std::invalid_argument(errorMessage);
   }
 }
+
--- a/libpimeval/src/pimParamsDDRDram.cpp
+++ b/libpimeval/src/pimParamsDDRDram.cpp
@@ -1,0 +1,154 @@
+// File: pimParamsDram.cc
+// PIMeval Simulator - DRAM parameters
+// Copyright (c) 2024 University of Virginia
+// This file is licensed under the MIT License.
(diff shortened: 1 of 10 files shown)


libpimeval/src/pimParamsDDRDram.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+    m_epochPeriod = std::stoi(pimUtils::getParam(params, "epoch_period"));
+    m_outputLevel = std::stoi(pimUtils::getParam(params, "output_level"));
+  } catch (const std::invalid_argument& e) {
+    std::string errorMessage("PIM-Error: Missing or invalid parameter: ");
+    errorMessage += e.what();
+    errorMessage += "\n";
+    throw std::invalid_argument(errorMessage);
+  }
+}
```
deyuan, line 154: There are warning signs for missing line break at the end of a few files

deyuan (approved):
Looks good to me. Suggest merge PR #172 first, and run the full testing before merging.

libpimeval/src/pimParamsDDRDram.cpp
fasiddique: Thanks! Updated!

PR #170, opened 2024-10-04
Adding FP32 support.

fzx333:
Still developing float support; no need to merge.


deyuan:
Once PR #172 is merged, we can enable the MUL_SCALAR testing in test-functional for coverage.

deyuan:
Suggest merging PR #175 first to handle broadcast, type casting, etc.

libpimeval/src/pimPerfEnergyTables.cpp
```diff
@@ -524,6 +525,7 @@ pimPerfEnergyTables::bitsimdPerfTable = {
       { PimCmdEnum::SUB,          { 1597,  822, 2024 } },
       { PimCmdEnum::MUL,          { 2222, 1200, 3664 } },
       { PimCmdEnum::DIV,          { 3292, 1749, 5024 } },
+      { PimCmdEnum::MUL_SCALAR,   { 1852, 1000, 3054 } },
```
deyuan, line 528: You might want to copy the numbers from line 526 here instead of line 269 which is for a different PIM variant

tests/test-functional/test-functional-fp.h
```diff
@@ -253,8 +255,9 @@ testFunctional::testFp(const std::string& category, PimDataType dataType)
           default: assert(0);
         }
         if (vecDest[i] != expected) {
+          // FP32 DIVISION sometimes(in extreme cases where float precision has limit) gives different results that have only very small difference compared to the number itself.
```
deyuan, line 258: I suspect this is because of different sequence of float/double type casting. Could you double check if there is any float/double casting introduced in this testing and during functional simulation?

tests/test-functional/test-functional-fp.h
```diff
@@ -76,37 +78,37 @@ testFunctional::testFp(const std::string& category, PimDataType dataType)
       {  1, "pimSub"                 },
       {  2, "pimMul"                 },
       {  3, "pimDiv"                 },
-      //{  4, "pimAbs"                 }, // TODO
+      {  4, "pimAbs"                 }, // TODO
```
deyuan, line 81: TODO can be removed

deyuan (approved):

PR #176, opened 2024-10-06
Removed the intrinsic header.

fzx333:
Hi, @deyuan; that header was used to call an architecture-specific prefetch instruction. It was a minor optimization, but it turns out that it has little impact on the performance, so I removed it for now. You can try to compile again and the code should work fine now!


deyuan (approved):
Great. The compilation issue is resolved. Thanks!

PR #189, opened 2024-10-17
FP32 functional support for all operation completed!

fzx333:
FP32 functional support completed, requesting a merge to the main.

Changes in this commit:

- Fixed the divscalar results inconsistent issue by changing -Ofast to -O3.
- Added FP32 support for all operations.
- Code cleaning up in pimCmdFunc2 (Replacing the big if-else block with a templete function).

### DIFF
--- a/libpimeval/src/libpimeval.h
+++ b/libpimeval/src/libpimeval.h
@@ -146,9 +146,11 @@ PimStatus pimScaledAdd(PimObjId src1, PimObjId src2, PimObjId dest, uint64_t sca
 PimStatus pimPopCount(PimObjId src, PimObjId dest);
 PimStatus pimRedSumInt(PimObjId src, int64_t* sum);
 PimStatus pimRedSumUInt(PimObjId src, uint64_t* sum);
+PimStatus pimRedSumFP32(PimObjId src, float* sum);
 // Note: Reduction sum range is [idxBegin, idxEnd)
 PimStatus pimRedSumRangedInt(PimObjId src, uint64_t idxBegin, uint64_t idxEnd, int64_t* sum);
 PimStatus pimRedSumRangedUInt(PimObjId src, uint64_t idxBegin, uint64_t idxEnd, uint64_t* sum);
+PimStatus pimRedSumRangedFP32(PimObjId src, uint64_t idxBegin, uint64_t idxEnd, float* sum);
 PimStatus pimBroadcastInt(PimObjId dest, int64_t value);
 PimStatus pimBroadcastUInt(PimObjId dest, uint64_t value);
 PimStatus pimBroadcastFP32(PimObjId dest, float value);
--- a/libpimeval/src/pimCmd.cpp
+++ b/libpimeval/src/pimCmd.cpp
@@ -598,7 +598,7 @@ pimCmdRedSum<T>::computeRegion(unsigned index)
     if (currIdx >= m_idxBegin) {
       uint64_t operandBits = objSrc.getElementBits(currIdx);
       bool isFP = (dataType == PIM_FP32);
-      if(!isFP){
+      if (!isFP) {
         T integerOperand = pimUtils::signExt(operandBits, dataType);
         m_regionSum[index] += integerOperand;
       } else if (isFP){
--- a/libpimeval/src/pimCmd.cpp
+++ b/libpimeval/src/pimCmd.cpp
@@ -407,7 +407,7 @@ pimCmdFunc1::computeRegion(unsigned index)
       if(!computeResultFP(floatOperand, m_cmdType, pimUtils::castBitsToType<float>(m_scalarValue), result)) return false;
       objDest.setElement(elemIdx, result);
     } else {
-    assert(0); // todo: data type
+      assert(0); // todo: data type
     }
   }
   return true;
--- a/libpimeval/src/pimCmd.cpp
+++ b/libpimeval/src/pimCmd.cpp
@@ -394,7 +394,7 @@ pimCmdFunc1::computeRegion(unsigned index)
         int64_t result = 0;
         if(!computeResult(signedOperand, m_cmdType, (int64_t)m_scalarValue, result, bitsPerElementSrc)) return false;
         objDest.setElement(elemIdx, result);
-      } else if (dataType != PIM_FP32) { // signed int.
+      } else if (dataType != PIM_FP32) { // unsigned int.
         uint64_t unsignedOperand = objSrc.getElementBits(elemIdx);
         uint64_t result = 0;
         if(!computeResult(unsignedOperand, m_cmdType, m_scalarValue, result, bitsPerElementSrc)) return false;
@@ -403,7 +403,7 @@ pimCmdFunc1::computeRegion(unsigned index)
         uint64_t bits = objSrc.getElementBits(elemIdx);
         float floatOperand = pimUtils::castBitsToType<float>(bits);
         float result = 0.0;
-        if(!computeResultFP(floatOperand, m_cmdType, pimUtils::castBitsToType<float>(m_scalarValue), result, bitsPerElementSrc)) return false;
+        if(!computeResultFP(floatOperand, m_cmdType, pimUtils::castBitsToType<float>(m_scalarValue), result)) return false;
         objDest.setElement(elemIdx, result);
       } else {
       assert(0); // todo: data type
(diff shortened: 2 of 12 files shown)


libpimeval/src/libpimeval.h
```diff
@@ -146,9 +146,11 @@ PimStatus pimScaledAdd(PimObjId src1, PimObjId src2, PimObjId dest, uint64_t sca
 PimStatus pimPopCount(PimObjId src, PimObjId dest);
 PimStatus pimRedSumInt(PimObjId src, int64_t* sum);
 PimStatus pimRedSumUInt(PimObjId src, uint64_t* sum);
+PimStatus pimRedSumFP32(PimObjId src, float* sum);
```
fasiddique, line 149: Are we planning to have separate reduction sum APIs for different float types?

libpimeval/src/pimCmd.cpp
```diff
@@ -394,7 +394,7 @@ pimCmdFunc1::computeRegion(unsigned index)
         int64_t result = 0;
         if(!computeResult(signedOperand, m_cmdType, (int64_t)m_scalarValue, result, bitsPerElementSrc)) return false;
         objDest.setElement(elemIdx, result);
-      } else if (dataType != PIM_FP32) { // signed int.
+      } else if (dataType != PIM_FP32) { // unsigned int.
```
deyuan, line 397: Suggest defining an isUnsigned bool variable above and use it here. This avoids modifying this condition when introducing a new FP type.

libpimeval/src/pimCmd.cpp
```diff
@@ -493,96 +493,33 @@ pimCmdFunc2::computeRegion(unsigned index)
   unsigned numElementsInRegion = src1Region.getNumElemInRegion();
   for (unsigned j = 0; j < numElementsInRegion; ++j) {
     uint64_t elemIdx = elemIdxBegin + j;
-
-    if (dataType == PIM_INT8 || dataType == PIM_INT16 || dataType == PIM_INT32 || dataType == PIM_INT64 || dataType == PIM_UINT8 || dataType == PIM_UINT16 || dataType == PIM_UINT32 || dataType == PIM_UINT64) {
-      uint64_t operandBits1 = objSrc1.getElementBits(elemIdx);
-      uint64_t operandBits2 = objSrc2.getElementBits(elemIdx);
-      // The following if-else block is the perfect example of where a Template would have been much more cleaner and efficient and less error prone
-      if (dataType == PIM_INT8 || dataType == PIM_INT16 || dataType == PIM_INT32 || dataType == PIM_INT64) {
+    if (dataType == PIM_INT8 || dataType == PIM_INT16 || dataType == PIM_INT32 || dataType == PIM_INT64 || dataType == PIM_UINT8 || dataType == PIM_UINT16 || dataType == PIM_UINT32 || dataType == PIM_UINT64 || dataType == PIM_FP32) {
```
deyuan, line 496: We can remove this if condition in func1/func2. Some bool variables such as isSigned, isUnsigned, isFP would be sufficient for catching unknown types. 
if (isSigned) {
} else if (isUnsigned) {
} else if (FP32) {
} else {
  assert
}

libpimeval/src/pimCmd.cpp
```diff
@@ -653,14 +590,23 @@ pimCmdRedSum<T>::computeRegion(unsigned index)
...
 
   unsigned numElementsInRegion = srcRegion.getNumElemInRegion();
   uint64_t currIdx = srcRegion.getElemIdxBegin();
   for (unsigned j = 0; j < numElementsInRegion && currIdx < m_idxEnd; ++j) {
     if (currIdx >= m_idxBegin) {
       uint64_t operandBits = objSrc.getElementBits(currIdx);
-      T operand = pimUtils::signExt(operandBits, objSrc.getDataType());
-      m_regionSum[index] += operand;
+      bool isFloat = (dataType == PIM_FP32);
+      if(!isFloat){
```
deyuan, line 601: Space after if and before {

tests/test-functional/test-functional-fp.h
```diff
@@ -204,6 +205,14 @@ testFunctional::testFp(const std::string& category, PimDataType dataType)
         sumUIntExpected += vecSrc1[i];
       }
       assert(sumUInt == sumUIntExpected);
+    }else if (testName == "pimRedSumFP32" || testName == "pimRedSumRangedFP32") {
```
deyuan, line 208: Space after }

tests/test-functional/test-functional-fp.h
```diff
@@ -78,37 +78,37 @@ testFunctional::testFp(const std::string& category, PimDataType dataType)
...
-      //{ 29, "pimRedSumInt"           }, // TODO
-      //{ 30, "pimRedSumUInt"          }, // TODO
-      //{ 31, "pimRedSumRangedInt"     }, // TODO
-      //{ 32, "pimRedSumRangedUInt"    }, // TODO
-      //{ 33, "pimBroadcastInt"        }, // TODO
-      //{ 34, "pimBroadcastFP32"       }, // TODO
+      //{ 29, "pimRedSumInt"           }, // not supported
+      { 30, "pimRedSumFP32"          },
+      //{ 31, "pimRedSumRangedInt"     }, // not supported
+      // { 32, "pimRedSumRangedFP32"    },
```
deyuan, line 109: Can this work?

deyuan (commented):
Added a few suggestions

libpimeval/src/libpimeval.h
deyuan: It depends on our short-term solution, e.g. using float, byte representation, or customized types to support FP8/16

fzx333:
Hi @deyuan, I have made those small code changes based on your suggestion. All the test cases are PASS except redSumRanged.

Now, the only issue left is the redSumRanged is not working, and the reason is probably because of that float compiler optimization issue again. Because when I print out the expected value and the value from PIMeval for redSumRanged, they only have a slight marginal difference. (If I reduce the range of the sum, sometimes it works with no error.) Also, the redSum is always working fine, so I don't think this is a bug because of the code functionality. I also tried -O1 and debug, but the error still exists. Can you try to test the code in your setup and see if you have the issue? Thanks!

libpimeval/src/pimCmd.cpp
```diff
@@ -653,14 +590,23 @@ pimCmdRedSum<T>::computeRegion(unsigned index)
...
 
   unsigned numElementsInRegion = srcRegion.getNumElemInRegion();
   uint64_t currIdx = srcRegion.getElemIdxBegin();
   for (unsigned j = 0; j < numElementsInRegion && currIdx < m_idxEnd; ++j) {
     if (currIdx >= m_idxBegin) {
       uint64_t operandBits = objSrc.getElementBits(currIdx);
-      T operand = pimUtils::signExt(operandBits, objSrc.getDataType());
-      m_regionSum[index] += operand;
+      bool isFP = (dataType == PIM_FP32);
+      if(!isFP){
```
deyuan, line 601:
```suggestion
      if (!isFP) {
```

libpimeval/src/pimCmd.cpp
```diff
@@ -653,14 +590,23 @@ pimCmdRedSum<T>::computeRegion(unsigned index)
...
+        T integerOperand = pimUtils::signExt(operandBits, dataType);
+        m_regionSum[index] += integerOperand;
+      } else if (isFP){
```
deyuan, line 604:
```suggestion
      } else if (isFP) {
```

libpimeval/src/pimCmd.cpp
```diff
@@ -387,27 +387,27 @@ pimCmdFunc1::computeRegion(unsigned index)
...
+      if(!computeResult(unsignedOperand, m_cmdType, m_scalarValue, result, bitsPerElementSrc)) return false;
+      objDest.setElement(elemIdx, result);
+    } else if (isFP){
+      uint64_t bits = objSrc.getElementBits(elemIdx);
+      float floatOperand = pimUtils::castBitsToType<float>(bits);
+      float result = 0.0;
+      if(!computeResultFP(floatOperand, m_cmdType, pimUtils::castBitsToType<float>(m_scalarValue), result)) return false;
+      objDest.setElement(elemIdx, result);
+    } else {
+    assert(0); // todo: data type
```
deyuan, line 410:
```suggestion
      assert(0); // todo: data type
```

tests/test-functional/test-functional-fp.h
```diff
@@ -45,7 +45,7 @@ testFunctional::testFp(const std::string& category, PimDataType dataType)
 
   // Cover scalar value testing
   // PIMeval uses uint64_t to represent bits of scalarValue
-  const float scalarValFloat = 123.0;
+  const float scalarValFloat = 123.0f;
   const uint64_t scalarVal = *reinterpret_cast<const uint64_t*>(&scalarValFloat);
```
deyuan, line 49:
```suggestion
  const uint64_t scalarVal = *reinterpret_cast<const uint32_t*>(&scalarValFloat);
```

deyuan: Float type has 32 bits only. Reinterpreting it with uint64_t* can lead to some incorrect results

deyuan:
> Hi @deyuan, I have made those small code changes based on your suggestion. All the test cases are PASS except redSumRanged.
> 
> Now, the only issue left is the redSumRanged is not working, and the reason is probably because of that float compiler optimization issue again. Because when I print out the expected value and the value from PIMeval for redSumRanged, they only have a slight marginal difference. (If I reduce the range of the sum, sometimes it works with no error.) Also, the redSum is always working fine, so I don't think this is a bug because of the code functionality. I also tried -O1 and debug, but the error still exists. Can you try to test the code in your setup and see if you have the issue? Thanks!

I tried your code branch and observed the same. It is because when adding up a vector of float numbers, the order matters. We use multithreaded reduction in pimCmd and a simple loop in this test function, which can cause some result differences. I don't see an easy way to ensure the float reduction sum results are identical between the two. We can consider doing an almostEqual check here with a small tolerance.

fzx333:
Thank you for your suggestions @deyuan. Everything is updated now.

deyuan (approved):
If everything looks good, please do not forget to update test-functional/result-golden.txt along with this PR. Thanks!

fzx333:
Thanks for the reminder, I have updated the golden results based on my local results, before I merge, can you also take a brief look and let me know if everything looks good to you?

deyuan:
> Thanks for the reminder, I have updated the golden results based on my local results, before I merge, can you also take a brief look on the golden results and let me know if everything looks good to you?

Hi @fzx333, thanks for checking with me. There are a few assertion failures, and we cannot update such failures to result-golden.txt. I committed a few changes in your branch to resolve them. 
- Issue 1: Performance model of FP32 reduction sum on bit-serial PIM triggered an assert(0)
- Issue 2: An absolute error tolerance is not enough for reduction sum. I implement a fuzzeEqualPercent function in test-functional to make sure the error is within a small percentage.
- Issue 3: I'm seeing small rounding issues in scaled_add perf results, as shown in my commit [5d7b865](https://github.com/UVA-LavaLab/PIMeval-PIMbench/pull/189/commits/5d7b865f3276b93ebc6848a1319bb893cf97d175). I updated numbers in result-golden.txt, but could you try again and see if you are seeing different numbers on your machine?

deyuan (approved):

fzx333:
> > Thanks for the reminder, I have updated the golden results based on my local results, before I merge, can you also take a brief look on the golden results and let me know if everything looks good to you?
> 
> Hi @fzx333, thanks for checking with me. There are a few assertion failures, and we cannot update such failures to result-golden.txt. I committed a few changes in your branch to resolve them.
> 
> * Issue 1: Performance model of FP32 reduction sum on bit-serial PIM triggered an assert(0)
> * Issue 2: An absolute error tolerance is not enough for reduction sum. I implement a fuzzeEqualPercent function in test-functional to make sure the error is within a small percentage.
> * Issue 3: I'm seeing small rounding issues in scaled_add perf results, as shown in my commit [5d7b865](https://github.com/UVA-LavaLab/PIMeval-PIMbench/pull/189/commits/5d7b865f3276b93ebc6848a1319bb893cf97d175). I updated numbers in result-golden.txt, but could you try again and see if you are seeing different numbers on your machine?

Hi @deyuan, I do still see small rounding issues in scaled_add perf results on my machine, besides that, everything else is the same here! Is there an easy fix to avoid that slight number differences? Or do you think it's okay for now to have that small differences in results?

deyuan:
@fzx333 Let's get this PR merged. For the add_scalar perf results rounding issue, we can follow up separately and make sure the results are consistent across platforms.

fzx333:
> @fzx333 Let's get this PR merged. For the add_scalar perf results rounding issue, we can follow up separately and make sure the results are consistent across platforms.

Understand!

PR #201, opened 2024-11-04
Histogram alternative

hugoabbot:
hist algorithm which uses ranged reduction sum and less equals scalar calls.

**Note: the folder name is just `histogram-alt`, instead of `cpp-histogram-alt` (should the rest of misc-bench be refactored?)**

### DIFF
--- a/misc-bench/histogram-alt/hist-alt.cpp
+++ b/misc-bench/histogram-alt/hist-alt.cpp
@@ -1,0 +1,284 @@
+// Test: C++ version of Histogram alternative
+// Copyright (c) 2024 University of Virginia
+// This file is licensed under the MIT License.
+// See the LICENSE file in the root of this repository for more details.
+
+#include <iostream>
+#include <vector>
+#include <getopt.h>
+#include <stdint.h>
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <sys/mman.h>
+#include <unistd.h>
+#include <cstring>
+#include <cassert>
+
+#include "../util.h"
+#include "libpimeval.h"
+
+using namespace std;
+
+#define NUMBINS 256 // RGB values at any given pixel can be a value 0 to 255 (inclusive)
+#define NUMCHANNELS 3 // Red, green, and blue color channels
+
+// Params ---------------------------------------------------------------------
+typedef struct Params
+{
+  char *configFile;
+  std::string inputFile;
+  bool shouldVerify;
+} Params;
+
+void usage()
+{
+  fprintf(stderr,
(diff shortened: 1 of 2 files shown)


fasiddique:
> **Note: the folder name is just `histogram-alt`, instead of `cpp-histogram-alt` (should the rest of misc-bench be refactored?)**
Thank you for following the updated format. Yes, we should refactor the rest of the directories in a separate PR.

misc-bench/histogram-alt/hist-alt.cpp
```diff
@@ -0,0 +1,284 @@
...
+#include <sys/stat.h>
+#include <sys/mman.h>
+#include <unistd.h>
+#include <cstring>
+#include <cassert>
+
+#include "../util.h"
+#include "libpimeval.h"
+
+using namespace std;
```
fasiddique, line 20:
```suggestion
```

fasiddique: Additionally the code needs to use `std::vector` or similar

misc-bench/histogram-alt/hist-alt.cpp
```diff
@@ -0,0 +1,284 @@
...
+      orderedImgData.push_back(imgData[i + j]);
+    }
+  }
+
+  uint64_t numCol = deviceProp.numColPerSubarray, numRow = deviceProp.numRowPerSubarray, 
+           numCore = deviceProp.numRanks * deviceProp.numBankPerRank * deviceProp.numSubarrayPerBank;
+  uint64_t totalAvailableBits = numCol * numRow * numCore;
+  uint64_t requiredBitsforImage = ((imgDataBytes / NUMCHANNELS * 8) + 8); // Using uint8_t instead of int, only require 8 bits
+  int numItr = std::ceil(static_cast<double> (requiredBitsforImage) / totalAvailableBits);
+  // std::cout << "Required iterations for image: " << numItr << std::endl;
```
fasiddique, line 193:
```suggestion
```

misc-bench/histogram-alt/hist-alt.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+  if (numItr == 1)
+  {
+    histogram(imgDataBytes, orderedImgData, redCount, greenCount, blueCount);
+  }
+  else
```
fasiddique, line 199: I think we can remove the conditional `if-else` and based on the `numItr`, we can run it in multiple pass

fasiddique (approved):

PR #203, opened 2024-12-02
Support PIMeval analysis mode for fast performance and energy analysis

deyuan:
**Feature**: PIMeval analysis mode, off by default experimental feature
**Purpose**: Fast performance and energy analysis, 2-10+x
**UI to enable**: Environment variable PIMEVAL\_ANALYSIS\_MODE=1 
**Impact**: Once enabled, PIMeval will skip data copying and PIM computation during simulation for fast runtime, while the performance and energy estimation results will be identically preserved.
**Limitation**: This mode works for result-independent PIM kernels. Result-dependent PIM kernel will not work because the PIM computation results are not usable. In addition, some applications need to guard some result verification code with pimIsAnalysisMode check.

### DIFF
--- a/libpimeval/src/libpimeval.cpp
+++ b/libpimeval/src/libpimeval.cpp
@@ -55,6 +55,13 @@ pimResetStats()
   pimSim::get()->resetStats();
 }
 
+//! @brief  Is analysis mode. Call this after device creation
+bool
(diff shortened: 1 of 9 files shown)


fasiddique (approved):
LGTM!

PR #207, opened 2025-01-22
Updated reduction API call

fasiddique:


hugoabbot (approved):
Everything works on my end!

PR #208, opened 2025-01-24
Fix histogram alternative to compile with new reduction sum

Arleee1:
During testing, I noticed that compiling the full library no longer worked due to updates to the reduction sums in histogram alt. I have fixed this below, now the whole repository compiles, and I have checked the histogram alt changes using the default inputs.

### DIFF
--- a/libpimeval/src/libpimeval.h
+++ b/libpimeval/src/libpimeval.h
@@ -150,7 +150,7 @@ PimStatus pimMaxScalar(PimObjId src, PimObjId dest, uint64_t scalarValue);
 // multiply src1 with scalarValue and add the multiplication result with src2. Save the result to dest. 
 PimStatus pimScaledAdd(PimObjId src1, PimObjId src2, PimObjId dest, uint64_t scalarValue);
 PimStatus pimPopCount(PimObjId src, PimObjId dest);
-PimStatus pimRedSum(PimObjId src, void* min, uint64_t idxBegin = 0, uint64_t idxEnd = 0);
+PimStatus pimRedSum(PimObjId src, void* sum, uint64_t idxBegin = 0, uint64_t idxEnd = 0);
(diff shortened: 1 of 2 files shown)


libpimeval/src/libpimeval.h
```diff
@@ -150,7 +150,7 @@ PimStatus pimMaxScalar(PimObjId src, PimObjId dest, uint64_t scalarValue);
 // multiply src1 with scalarValue and add the multiplication result with src2. Save the result to dest. 
 PimStatus pimScaledAdd(PimObjId src1, PimObjId src2, PimObjId dest, uint64_t scalarValue);
 PimStatus pimPopCount(PimObjId src, PimObjId dest);
-PimStatus pimRedSum(PimObjId src, void* min, uint64_t idxBegin = 0, uint64_t idxEnd = 0);
+PimStatus pimRedSum(PimObjId src, void* sum, uint64_t idxBegin = 0, uint64_t idxEnd = 0);
```
fasiddique, line 153: This needs to be changed in the `libpimeval.cpp` file as well

fasiddique:
> I think you might have an outdated version of the hist-alt.cpp. The current main branch compiles for me and uses `static_cast<void*>(&____count[i])`. The change from `min` to `sum` would be helpful though.

I agree! 
@Arleee1 Thank you! Could you please rebase this with the main branch and also address the comment.

Arleee1:
> > I think you might have an outdated version of the hist-alt.cpp. The current main branch compiles for me and uses `static_cast<void*>(&____count[i])`. The change from `min` to `sum` would be helpful though.
> 
> I agree! @Arleee1 Thank you! Could you please rebase this with the main branch and also address the comment.

@hugoabbot I sincerely apologize, I was on an old commit. I have opened a new PR for the red sum rename and will close this one.

PR #209, opened 2025-01-24
Create String Matching Benchmark

Arleee1:


misc-bench/cpp-string-match/PIM/string-match.cpp
Arleee1, line 1: @fasiddique @deyuan I wanted to ask about this part of the code, to run the string matching benchmark as efficiently as possible, I need to allocAssociate as many objects with each other as possible. How can I best do this?

fasiddique: @Arleee1 you can use [pimGetDeviceProperties(PimDeviceProperties* deviceProperties)](https://github.com/UVA-LavaLab/PIMeval-PIMbench/blob/main/libpimeval/src/libpimeval.h#L103) API to check the capacity of the PIM device and this will give you the number for max possible `allocAssociated` object that can be created. While we currently do not have the example of exact use case but [here](https://github.com/UVA-LavaLab/PIMeval-PIMbench/blob/main/misc-bench/histogram-alt/hist-alt.cpp#L169) is an example that uses the API to calculate the capacity.

Arleee1: @fasiddique Will the max number of objects be different if the device is vertical or horizontal?

fasiddique: yes, that is correct! This will vary horizontal vs vertical data layout.

Arleee1:
> @Arleee1 please let me know if this PR is ready to review

@fasiddique  I think it is ready now, I have gotten the chance to further optimize the PIM variant, and to clean up the code slightly.

misc-bench/cpp-string-match/PIM/string-match.cpp
```diff
@@ -0,0 +1,314 @@
...
+          "\n    -t    text input file to search for keys from (required, searches in cpp-string-match/dataset directory)"
+          "\n    -c    dramsim config file"
+          "\n    -v    t = verifies PIM output with host output. (default=false)"
+          "\n");
+}
+
+struct Params getInputParams(int argc, char **argv)
+{
+  struct Params p;
+  p.keysInputFile = nullptr;
```
fasiddique, line 45: Can we please have a default input file?

misc-bench/cpp-string-match/PIM/string-match.cpp
```diff
@@ -0,0 +1,314 @@
...
+  p.textInputFile = nullptr;
```
fasiddique, line 46: A default input file needed

misc-bench/cpp-string-match/PIM/string-match.cpp
```diff
@@ -0,0 +1,314 @@
...
+    return 1;
+  }
+  
+  std::cout << "Running PIM string match for \"" << params.keysInputFile << "\" as the keys file, and \"" << params.textInputFile << "\" as the text input file\n";
+  
+  std::string haystack;
+  std::vector<std::string> needles;
+  std::vector<int> matches;
+
+  const std::string DATASET_FOLDER_PREFIX = "./../dataset/";
```
fasiddique, line 258: Also, I don't think we should make this mandatory to have the input file in the dataset folder. User should be able to run the application using input stored in any folder

fasiddique:
@Arleee1 can we please use camel case for function/variable names?

Arleee1:
> @Arleee1 can we please use camel case for function/variable names?

Done

misc-bench/cpp-string-match/PIM/string-match.cpp
```diff
@@ -0,0 +1,334 @@
...
+  
+  PimObjId intermediatePim = pimAllocAssociated(haystackPim, PIM_UINT32);
+  assert(intermediatePim != -1);
+
+  // TODO remove when pim conversion becomes possible
+  uint32_t *haystack32Bit = (uint32_t*) malloc(sizeof(uint32_t) * haystack.size());
+  #pragma omp parallel for
+  for(uint32_t i=0; i<haystack.size(); ++i) {
+    haystack32Bit[i] = (uint32_t) haystack[i];
+  }
```
fasiddique, line 155: This code block needs documentation. Could you please add some comments.

misc-bench/cpp-string-match/PIM/string-match.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+  std::vector<PimObjId> pimIndividualNeedleMatches;
+  for(uint32_t i=0; i<needlesTable[0][0].size(); ++i) {
```
fasiddique, line 158: This will crash if `needlesTable` is empty. I think it would be good to do a check if the vector is empty before accessing it.

misc-bench/cpp-string-match/PIM/string-match.cpp
```diff
@@ -0,0 +1,334 @@
...
+    pimIndividualNeedleMatches.push_back(pimAllocAssociated(haystackPim, PIM_UINT32));
+    assert(pimIndividualNeedleMatches.back() != -1);
```
fasiddique, line 160: What could be the maximum size of `needlesTable[0][0]`?

misc-bench/cpp-string-match/PIM/string-match.cpp
```diff
@@ -0,0 +1,334 @@
...
+  }
+
+  uint64_t needlesDone = 0;
+
+  pimStartTimer();
+
+  PimStatus status;
+  for(uint64_t iter=0; iter<needlesTable.size(); ++iter) {
+    status = pimCopyHostToDevice((void *)haystack32Bit, haystackPim);
```
fasiddique, line 169: Is `haystack32Bit` mutable inside the for loop? If not, this should be moved out of the for loop

misc-bench/cpp-string-match/PIM/string-match.cpp
```diff
@@ -0,0 +1,334 @@
...
+        status = pimMulScalar(pimIndividualNeedleMatches[needleIdxPim], pimIndividualNeedleMatches[needleIdxPim], 1 + currentNeedleIdx);
+        assert (status == PIM_OK);
+      } else {
+        status = pimXorScalar(pimIndividualNeedleMatches[needleIdxPim], pimIndividualNeedleMatches[needleIdxPim], 1);
+        assert (status == PIM_OK);
+
+        status = pimSubScalar(pimIndividualNeedleMatches[needleIdxPim], pimIndividualNeedleMatches[needleIdxPim], 1);
+        assert (status == PIM_OK);
+
+        status = pimAndScalar(pimIndividualNeedleMatches[needleIdxPim], pimIndividualNeedleMatches[needleIdxPim], 1 + currentNeedleIdx);
```
fasiddique, line 220: This block needs documentation. As in, why the PIM logic changes based on the data layout

misc-bench/cpp-string-match/PIM/string-match.cpp
Arleee1: The haystackPim object is shifted in the loop, but needs to be reset for each iteration of the outer loop. The pimCopyHostToDevice call currently accomplishes this, however this could be more efficiently replaced by keeping an unshifted copy on the device and calling pimCopyObjectToObject, once that API is available. I have added comments to the code on this, and plan to update it when pimCopyObjectToObject is available.

misc-bench/cpp-string-match/PIM/string-match.cpp
Arleee1: As the output is formatted as the 32 bit index of the needle for matches, there cannot be more than INT_MAX needles. Therefore, we can enumerate through the needles using a uint32_t variable

fasiddique: In this case, can you please reserve the vector and add objects by indexing instead of using `push_back` as that would be faster

misc-bench/cpp-string-match/PIM/string-match.cpp
fasiddique: @mhgholamrezaei could you please review and merge PR #220 so that it is available to use?

misc-bench/cpp-string-match/PIM/string-match.cpp
Arleee1: The two blocks accomplish the same thing, however the multiplication is faster for the bit parallel architectures I have tested (Fulcrum and Bank Level), while the binary manipulation is faster for the bit serial architecture (BitSIMD). I use isHorizontal as a way to check for which the device is. I have also explained this further in the comments.

misc-bench/cpp-string-match/PIM/string-match.cpp
Arleee1: Done

misc-bench/string-match/cpp-string-match/PIM/string-match.cpp
```diff
@@ -0,0 +1,400 @@
...
+  // Therefore, all calculations must be done in 32 bits, so we cast everything to 32 bits
+  // This can be removed when the simulator supports mixed width operands
+  // TODO: remove when pim conversion becomes possible
+  uint32_t *haystack32Bit = (uint32_t*) malloc(sizeof(uint32_t) * haystack.size());
+  #pragma omp parallel for
+  for(uint32_t i=0; i<haystack.size(); ++i) {
+    haystack32Bit[i] = (uint32_t) haystack[i];
+  }
+
+  if(needlesTable.size() == 0 || needlesTable[0].size() == 0) {
```
fasiddique, line 164:
```suggestion
  if(needlesTable.empty() || needlesTable[0].empty()) {
```

misc-bench/string-match/PIM/slurm.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+#SBATCH --nodes=1
+#SBATCH --ntasks=1
+#SBATCH --partition=cpu
+#SBATCH --cpus-per-task=60
+#SBATCH --mem=100G
+#SBATCH --time=4-00:00:00
+#SBATCH --mail-type=all
+#SBATCH --mail-user=keg9ve@virginia.edu
```
fasiddique, line 9: Consider removing your email ID to avoid receiving unintended notifications if others run this, unless you are a fan of spams!

misc-bench/string-match/PIM/run_slurm.sh
fasiddique, line 1: Could you please let me know what is the purpose of this file?

Arleee1: I just used it for convenience when testing, it isn't necessary, I deleted it.

misc-bench/string-match/PIM/slurm.sh
Arleee1: Done

misc-bench/string-match/PIM/slurm.sh
```diff
@@ -0,0 +1,25 @@
...
+
+test_dirs=("$@")
+
+pim_arch_config_prefix="./../../../configs/iiswc"
+pim_arch_configs=("PIMeval_Bank_Rank32.cfg" "PIMeval_BitSerial_Rank32.cfg" "PIMeval_Fulcrum_Rank32.cfg")
+
+for test_dir in "${test_dirs[@]}"; do
```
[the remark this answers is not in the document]
Arleee1: The run_slurm file would put them into files "pim-text_len.log", "pim-num_keys.log", and "pim-key_len.log", based on the test that was being done, should I add that back?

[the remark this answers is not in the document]
Arleee1: Also, I believe that slurm will automatically write stdout and stderr to log files by default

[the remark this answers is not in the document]
fasiddique: No that is ok. I am also wondering if we should keep the slurm script. For now, let's keep it

fasiddique (approved):
Looks good to me! I have cloned and ran this for PIM and it works fine. PS we need to come back after mixed data type is supported. Thank you! Please feel free to merge it!

PR #210, opened 2025-01-24
Rename min to sum for pimRedSum parameter

Arleee1:
The parameter in libpimeval.cpp is already named sum, but it is currently named min in libpimeval.h


fasiddique (approved):
Thank you!

PR #211, opened 2025-01-29
Measure CPU/PIM portion of kernel runtime

deyuan:
Purpose: Measure CPU/PIM portion of kernel runtime for modeling refresh. CPU runtime is based on real elapsed time. PIM runtime is based on estimated runtime.
Changes:
- Add pimStartTimer and pimEndTimer APIs to mark boundary of kernels and track CPU and PIM runtime separately
- Show stats when pimEndTimer is called

### DIFF
--- a/libpimeval/src/pimStats.h
+++ b/libpimeval/src/pimStats.h
@@ -37,41 +37,22 @@ public:
   pimStatsMgr() {}
   ~pimStatsMgr() {}
 
+  void startKernelTimer();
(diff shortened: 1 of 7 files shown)


libpimeval/src/pimStats.h
fasiddique, line 1: This clean up is great! Thank you!

fasiddique:
Thank you!

fasiddique (approved):

PR #213, opened 2025-01-31
Hyperdimensional Computing (HDC) Benchmark

mhgholamrezaei:
Changes:
- Hyperdimensional Computing (HDC) pytorch and PIM implementations are added.


PIMbench/hdc/PIM/hdc.cpp
```diff
@@ -0,0 +1,350 @@
...
+        throw std::runtime_error("Could not open file");
+    }
+
+    std::string line;
+    if (std::getline(file, line))
+    {
+        std::istringstream dim_stream(line);
+        int dim1, dim2;
+        char comma;
+        if (!(dim_stream >> dim1 >> comma >> dim2))
```
fasiddique, line 112: Could you please explain the purpose of this?

mhgholamrezaei: CSV files are used to store the tensors. The dimensions of the tensor is stored in the first line of the CSV file. This code snippet reads the dimensions of the tensor from the CSV files. Please, let me know if you have further questions.

fasiddique: Thank you! It is indeed a great idea to check if the format is correct. Would it be less error-prone to merge this with the while loop line#118 where parsing happens? Since we need to check all lines except the first one?

PIMbench/hdc/README.md
fasiddique, line 1: Should the UCSD GitHub repo be mentioned in the README as the reference implementation that this version is based on?

PIMbench/hdc/PIM/hdc.cpp
mhgholamrezaei: The first line of the CSV contains two elements, while the other lines have variable lengths. I believe we can merge these two sections of the function, but it may reduce readability. If your main concern is minimizing the number of lines in the function, I can split it into two separate functions instead.

PIMbench/hdc/README.md
mhgholamrezaei: Thats a good question, but, I am not sure about the answer.

fasiddique: If it is public then I think we should add it including any papers from them so that people can cite their paper if they want!

PIMbench/hdc/PIM/hdc.cpp
fasiddique: In that case let's keep it this way!

PIMbench/hdc/README.md
mhgholamrezaei: We should definitely cite the related works in the paper. Should we also include them in the README?

fasiddique: Yes, please!

fasiddique (approved):
Thank you! Please merge it after the README has been updated!

Issue #19, opened 2024-07-23
Update performance modeling of the last region for Fulcrum/bank-level


deyuan:
The last region has fewer elements which needs less time

PR #218, opened 2025-02-10
Fulcrum Reduction Updated With Even Load Balancing

fasiddique:
Related Issue #19


alifahmed:
It would be great if the core count & CPU frequency were not hardcoded. But for now, this looks good! Feel free to merge this

alifahmed (approved):

PR #220, opened 2025-02-14
Copy between objects

fasiddique:
Closes issue #124


tests/test-functional/test.txt
mhgholamrezaei, line 1: Should we have the output log in the codebase?

libpimeval/src/pimPerfEnergyBankLevel.cpp
```diff
@@ -23,6 +23,16 @@ pimPerfEnergyBankLevel::getPerfEnergyForFunc1(PimCmdEnum cmdType, const pimObjIn
...
+    case PimCmdEnum::COPY_O2O:
+    {
+      // How many iteration require to read / write max elements per region
+      unsigned numGDLItr = maxElementsPerRegion * bitsPerElement / m_GDLWidth;
+      double totalGDLOverhead = m_tGDL * numGDLItr; // read can be pipelined and write cannot be pipelined
+      msRuntime = (m_tR + m_tW + totalGDLOverhead) * numPass;
+      mjEnergy = numPass * numCores * ((m_eAP * 2) + (m_eGDL * 2));
+      mjEnergy += m_pBChip * m_numChipsPerRank * m_numRanks * msRuntime;
+      break;
+    }
```
mhgholamrezaei, line 35: Are the data be assumed to be copied over GDL? Why cannot the data be copied inside the sub-array similar to the Fulcrum architecture?

mhgholamrezaei (changes requested):
I made some comments. Please check them out.

tests/test-functional/test.txt
fasiddique: Good catch! This needs to be removed

libpimeval/src/pimPerfEnergyBankLevel.cpp
fasiddique: For bank-level PIM, we  do not have walkers attached to each subarray. As a result we need to fetch the data to the bank interface and then write it back. Please let me know if there's another way to do that without added walkers/registers

mhgholamrezaei: If this is our assumption, then everything should be fine.

PR #224, opened 2025-02-19
Fix a compilation error in string match for OSX.

deyuan:
Mismatched data types when calling std::max

### DIFF
--- a/misc-bench/string-match/include/string-match-utils.h
+++ b/misc-bench/string-match/include/string-match-utils.h
@@ -6,10 +6,10 @@
 #ifndef STRING_MATCH_UTILS_H
 #define STRING_MATCH_UTILS_H
 
-#include <vector>                      // for vector
-#include <string>                      // for string
-#include <fstream>                     // for ifstream
-#include <sstream>                     // for ostringstream
+#include <vector>
+#include <string>
+#include <fstream>
+#include <sstream>
 #include <algorithm>                   // for max
 #include <limits>                      // for numeric_limits
 
--- a/misc-bench/string-match/include/string-match-utils.h
+++ b/misc-bench/string-match/include/string-match-utils.h
@@ -1,10 +1,18 @@
+// Test: String matching utilities
+// Copyright (c) 2024 University of Virginia
+// This file is licensed under the MIT License.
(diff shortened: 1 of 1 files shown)


misc-bench/string-match/include/string-match-utils.h
```diff
@@ -1,10 +1,18 @@
...
-#include <string>
-#include <fstream>
-#include <sstream>
+#include <vector>                      // for vector
+#include <string>                      // for string
+#include <fstream>                     // for ifstream
+#include <sstream>                     // for ostringstream
+#include <algorithm>                   // for max
+#include <limits>                      // for numeric_limits
+
```
fasiddique, line 15: Please explain why this is necessary

deyuan: Although it looks straightforward for a few STL containers, the inline comments are very helpful in general. One can easily search for the standard APIs being used, and notice any unused headers once code is change. Except string and vector comments are added for consistency, you can see other APIs are not very self explanatory. I recommend to keep this formatting for standard C++ libraries. Could you let me know if there is any further concern?

fasiddique: I don’t believe inline comments for well-known standard libraries are helpful. They just introduce noise and redundancy. As I’ve mentioned earlier, if a particular library is used in non-trivial way or requires clarification, a well-placed comment is absolutely necessary. However, adding comments to every header doesn’t seem justified. Additionally, this approach is not a common practice in open-source projects, such as LLVM. If you have specific concerns beyond personal preference, please provide examples.

deyuan: @fasiddique Thanks for the review comments. As I've seen many pains and pitfalls for large scale of collaboration, I'm willing to improve the code quality and make things simpler and more explicit. Let's take this file as an example. Copyright is missing, data types for std::max are not correct, "algorithm" and "limits" headers are not there. The APIs are different from "fstream" and "sstream". Your concern is for the redundant "vector" and "string" comments, which I can revert. If you look at the headers in libpimeval, who will track what are being used and what are not? Do you agree that my changes can ease future tracking?

fasiddique (approved):

PR #225, opened 2025-02-19
IWYU: include-what-you-use changes for libpimeval module

deyuan:
Add and remove some standard C++ header files with clear comments at column 40.

### DIFF
--- a/libpimeval/src/pimUtils.cpp
+++ b/libpimeval/src/pimUtils.cpp
@@ -6,14 +6,22 @@
 
 #include "pimUtils.h"
 #include "libpimeval.h"
-#include <fstream>
-#include <iostream>
-#include <sstream>
-#include <unordered_map>
-#include <string>
-#include <filesystem>
-#include <cstdlib>
-#include <cassert>
+#include <fstream>                     // for ifstream
+#include <iostream>                    // for cerr, endl
+#include <sstream>                     // for stringstream, istringstream
+#include <unordered_map>               // for unordered_map
+#include <string>                      // for string, getline
+#include <filesystem>                  // for filesystem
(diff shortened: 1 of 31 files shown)


deyuan:
> C++ headers are self-explanatory, so I don’t see the need for adding comments to each one. Could you clarify the reasoning behind this? If there are unused or redundant headers, I agree we should remove them, but in this PR, I only see comments added.

Many standard C++ headers are not self-explanatory. Although it looks straightforward for a few STL containers, the inline comments are very helpful in general. One can easily search for the standard APIs being used, see why a header is included, and notice any unused headers once code are changed. I strongly recommend following this formatting for standard C++ libraries. Could you let me know if there is any further concern?

libpimeval/src/pimUtils.cpp
```diff
@@ -6,14 +6,22 @@
...
+#include <filesystem>                  // for filesystem
+#include <cstdlib>                     // for printf, getenv
+#include <cassert>                     // for assert
+#include <cctype>                      // for isspace
+#include <algorithm>                   // for find_if
+#include <memory>                      // for unique_lock
+#include <mutex>                       // for mutex
+#include <exception>                   // for exception
+#include <stdexcept>                   // for out_of_range, invalid_argument
+#include <limits>                      // for numeric_limits
```
deyuan, line 23: With these inline comments, we can easily see why a standard C++ header is included. There are many unobvious ones. The obvious ones are for consistency.

deyuan (commented):
Add a comment

deyuan:
Hi @fasiddique please help to review and do let me know if any changes fall into the "noise" category in your opinion. This PR is not only for header comments changes, but it adds required ones and removes used ones.

PR #231, opened 2025-02-21
Load-balance can be turned on/off using config file parameters **suggested by Kevin**

fasiddique:


libpimeval/src/pimResMgr.cpp
```diff
@@ -69,6 +69,7 @@ pimObjInfo::finalize()
   }
   m_numCoresUsed = coreIdCnt.size();
   m_numCoreAvailable = m_device->getNumCores();
+  m_isLoadBalanced = m_device->isLoadBalanced();
```
alifahmed, line 72: while I think this is fine for now, in the future (when there's no imminent deadline), a good improvement would be to return an immutable device with properties

libpimeval/src/pimResMgr.h
```diff
@@ -252,6 +253,7 @@ class pimObjInfo
   unsigned m_numColsPerElem = 0; // number of cols per element
   bool m_isDualContactRef = false;
   pimDevice* m_device = nullptr; // for accessing simulated memory
+  bool m_isLoadBalanced = true;
```
alifahmed, line 256: It's nothing major, but the default value might be set to false

alifahmed (approved):
lgtm. This is a helpful feature. Thanks for adding this! 
Aquabolt needs to be modified
Feel free to merge

PR #236, opened 2025-02-22
Support simulation target in device property API.

deyuan:
Main changes:
- Add simTarget field to PimDeviceProperties struct
- Update regression test

Per request from @mhgholamrezaei : 
During functional simulation, we would like to query the simulation target in application code.

No impact on existing behavior.


fasiddique (approved):
lgtm!

PR #238, opened 2025-02-24
Hamming string match

Arleee1:


Arleee1:
PR is not yet ready for review, I have discovered a corner case I need to account for.

misc-bench/hamming-string-match/hamming-data-generator/hamming-data-generator.cpp
fasiddique, line 1: @Arleee1 do you think this and the previous string match generator can be in the same file? Should we have one file for all the string match data genrator?

Arleee1: Good idea, where should I put the singular file?

misc-bench/hamming-string-match/PIM/hamming-string-match.cpp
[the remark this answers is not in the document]
Arleee1: I do not believe it should have the same error, however I cannot directly test on mac, as I do not have access to one.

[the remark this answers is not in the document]
fasiddique: No problem! I will test it once this is ready to merge

misc-bench/hamming-string-match/PIM/hamming-string-match.cpp
```diff
@@ -0,0 +1,561 @@
...
+
+  resultTableActualToSorted.resize(needles.size());
+  for(uint64_t needleIdx=0; needleIdx<needles.size(); ++needleIdx) {
+    resultTableActualToSorted[sortedToActualNeedles[needleIdx]] = needleIdx;
+  }
+  
+  // If vertical, each pim object takes 32 rows, 1 row if horizontal
+  // Three objects used by haystack, intermediate, and haystack copy
+  // Haystack copy only needed if more than one iteration
+  uint64_t maxNeedlesPerIteration = (isHorizontal ? numRows : (numRows>>5)) - 3;
```
fasiddique, line 156: This will not always be 5 right? It depends on the type? 5 is used for 2^5? if this is the case it would be great to make it a variable

misc-bench/hamming-string-match/PIM/hamming-string-match.cpp
```diff
@@ -0,0 +1,561 @@
...
+
+//! @brief  Precomputes a more optimal order to match keys/needles on PIM device for calculation reuse
+//! @details Instead of calling pimNEScalar multiple times for the same character, call only once per character per char index and reuse result
+//!          Orders needles within a character index to place needles with the same character next to each other
+//!          Not included in benchmarking time, as it only depends on the keys/needles and not on the text/haystack
+//! @param[in]  needles  A list of needles that will be matched
+//! @param[in]  numRows  The number of rows in the PIM device, needed to find the max number of needles per iteration
+//! @param[in]  isHorizontal  Represents if the PIM device is horizontal, needed to find the max number of needles per iteration
+//! @return  A table representing a better ordering to match the needles
+NeedlesTable stringMatchPrecomputeTable(const std::vector<std::string>& needles, const uint64_t numRows, const bool isHorizontal) {
```
fasiddique, line 135: Is this a one time cost? Are we including this in PIM performance?

Arleee1: It is a one time cost that can be reused to match the same keys across multiple texts, so I did not plan on including it in the benchmarking time. This is in line with with all three architectures from the other string matching benchmark (CPU, GPU, PIM), which also have one time costs that are not included.

misc-bench/hamming-string-match/hamming-data-generator/hamming-data-generator.cpp
Arleee1: I have moved most of the functionality to the utilStringMatch.h header, while keeping both of the data generator source files in the same place. I believe this way makes the most sense, so that both string matching benchmarks can have their own generators, without repeating code.

fasiddique (approved):

PR #240, opened 2025-02-24
Add names for ne and ne_scalar

Arleee1:
NE and NE Scalar currently show as "unknown" in the list of PIM commands, this resolves that.


deyuan (approved):
Hi @Arleee1 , thanks for fixing this! This is a newly added feature, and I haven't had a chance to add it to test-functional yet.

PR #242, opened 2025-02-26
PIM stat shows for each API percent time spend in row read, row write and compute 

fasiddique:
Aquabolt to be updated


libpimeval/src/pimStats.cpp
```diff
@@ -122,17 +122,26 @@ void
...
+    double percentRead = (it.second.second.m_msRead * 100 / it.second.second.m_msRuntime);
+    double percentWrite = (it.second.second.m_msWrite * 100 / it.second.second.m_msRuntime);
+    double percentCompute = (it.second.second.m_msCompute * 100 / it.second.second.m_msRuntime);
+    std::printf(" %44s : %10d %14f %14f %14.2f %14.2f %14.2f\n", it.first.c_str(), it.second.first, it.second.second.m_msRuntime, it.second.second.m_mjEnergy, percentRead, percentWrite, percentCompute);
     totalCmd += it.second.first;
     totalMsRuntime += it.second.second.m_msRuntime;
     totalMjEnergy += it.second.second.m_mjEnergy;
+    totalMsRead += it.second.first * percentRead;
+    totalMsWrite += it.second.first * percentWrite;
+    totalMsCompute += it.second.first * percentCompute;
```
fasiddique, line 142: This is for weighted average

libpimeval/src/pimPerfEnergyAquabolt.cpp
alifahmed, line 1: If you want to be consistent with the other architectures, variable names here need to be changed. You could also change those whenever you update them with actual value. I am fine with either.

libpimeval/src/pimStats.cpp
```diff
@@ -122,17 +122,26 @@ void
...
   double totalMjEnergy = 0.0;
+  double totalMsRead = 0.0;
+  double totalMsWrite = 0.0;
+  double totalMsCompute = 0.0;
   for (const auto& it : m_cmdPerf) {
-    std::printf(" %44s : %10d %14f %14f\n", it.first.c_str(), it.second.first, it.second.second.m_msRuntime, it.second.second.m_mjEnergy);
+    double percentRead = (it.second.second.m_msRead * 100 / it.second.second.m_msRuntime);
+    double percentWrite = (it.second.second.m_msWrite * 100 / it.second.second.m_msRuntime);
+    double percentCompute = (it.second.second.m_msCompute * 100 / it.second.second.m_msRuntime);
+    std::printf(" %44s : %10d %14f %14f %14.2f %14.2f %14.2f\n", it.first.c_str(), it.second.first, it.second.second.m_msRuntime, it.second.second.m_mjEnergy, percentRead, percentWrite, percentCompute);
```
deyuan, line 136: Suggest doing 6.2f or 7.2f for percentage columns. 14.2f is too wide

libpimeval/src/pimPerfEnergyAquabolt.cpp
deyuan: Agreed. It's better to use ms but not percent as variable prefix

fasiddique:
@alifahmed @deyuan thanks for the comments! Updated

alifahmed (approved):

PR #246, opened 2025-03-04
Re-golden test-functional with the latest main branch.

deyuan:
Besides column width changes, I see that bank level performance and energy numbers are also changed, which might be because of recent changes. @fasiddique could you please review and confirm? Thanks


fasiddique (approved):

fasiddique:
> Besides column width changes, I see that bank level performance and energy numbers are also changed, which might be because of recent changes. @fasiddique could you please review and confirm? Thanks

Yes, bank-level defaults to load-balanced. So the numbers don't match with previous one

PR #263, opened 2025-03-28
Performance per watt stat

fasiddique:
**Performance Per Watt** is defined as:

```
Performance Per Watt = Total Operations / Total Energy (in Joules)
Total Ops = Vector Size × Number of Operation 
// For Pop Count, Number of Operation, Fulcrum = 12; Bank = 1; Bit-Serial = 1
// For Scaled Add, Number of Operation, Fulcrum = Bank = Bit-Serial = 2 
```


fasiddique:
> **Performance Per Watt** is defined as:
> 
> ```
> Performance Per Watt = Total Operations / Total Energy (in Joules)
> Total Ops = Vector Size × Number of Operation 
> // For Pop Count, Number of Operation, Fulcrum = 12; Bank = 1; Bit-Serial = 1
> // For Scaled Add, Number of Operation, Fulcrum = Bank = Bit-Serial = 2 
> ```

@deyuan please let me know if you have any concerns regarding bit-serial total op calculation formula

libpimeval/src/pimPerfEnergyBitSerial.cpp
```diff
@@ -160,7 +166,7 @@ pimPerfEnergyBitSerial::getPerfEnergyBitSerial(PimDeviceEnum deviceType, PimCmdE
   msLogic *= numPass;
   msRuntime *= numPass;
   mjEnergy *= numPass;
-  return pimeval::perfEnergy(msRuntime, mjEnergy, msRead, msWrite, msLogic);
+  return pimeval::perfEnergy(msRuntime, mjEnergy, msRead, msWrite, msLogic, totalOp);
```
deyuan, line 169: Could you see if you can compute totalOp at the end of this function right before this line? I see some switch/case may not be covered right now, and you probably need to time totalOp by numPass

libpimeval/src/pimStats.cpp
```diff
@@ -141,9 +142,10 @@ pimStatsMgr::showCmdStats() const
     totalMsRead += it.second.first * percentRead;
     totalMsWrite += it.second.first * percentWrite;
     totalMsCompute += it.second.first * percentCompute;
+    totalOp += it.second.second.m_totalOp;
   }
   std::printf(" %44s : %10d %14f %14f %7.2f %7.2f %7.2f\n", "TOTAL ---------", totalCmd, totalMsRuntime, totalMjEnergy, (totalMsRead / totalCmd), (totalMsWrite / totalCmd), (totalMsCompute / totalCmd) );
-
+  std::printf("\nPerformance Per Watt: %14f GOPS/W\n", totalOp * 1.0 / totalMjEnergy * 1e-6);
```
deyuan, line 148: Maybe consider aligning this with other outputs

deyuan:
> > **Performance Per Watt** is defined as:
> > ```
> > Performance Per Watt = Total Operations / Total Energy (in Joules)
> > Total Ops = Vector Size × Number of Operation 
> > // For Pop Count, Number of Operation, Fulcrum = 12; Bank = 1; Bit-Serial = 1
> > // For Scaled Add, Number of Operation, Fulcrum = Bank = Bit-Serial = 2 
> > ```
> 
> @deyuan please let me know if you have any concerns regarding bit-serial total op calculation formula

Do you think if we can treat popcount per element as one OP for Fulcrum? Those 12 operations per popcount on Fulcrum are similar like bit-serial micro-programs IMO, which are internal details.

libpimeval/src/pimPerfEnergyBitSerial.cpp
fasiddique: For command such as `BIT_SLICE_INSERT` and `COPY_O2O` total OP should be zero for bit-serial. As a result, we cannot calculate total OP right before this like. Some of the switch/case were ignored for the same reason. But please let me know if I am mistaken.

libpimeval/src/pimPerfEnergyBankLevel.cpp
```diff
@@ -75,6 +76,7 @@ pimPerfEnergyBankLevel::getPerfEnergyForFunc1(PimCmdEnum cmdType, const pimObjIn
       mjEnergy += (m_eR * maxGDLItr * (numPass-1) + (m_eR * minGDLItr));
       mjEnergy += (m_eW * maxGDLItr * (numPass-1) + (m_eW * minGDLItr));
       mjEnergy += m_pBChip * m_numChipsPerRank * m_numRanks * msRuntime;
+      totalOp = obj.getNumElements();
```
fasiddique, line 79: should type conversion be considered an operation?

libpimeval/src/pimPerfEnergyBankLevel.h
```diff
@@ -31,9 +31,10 @@ class pimPerfEnergyBankLevel : public pimPerfEnergyBase
 protected:
   double m_blimpCoreLatency = 0.000005; // ms; 200 MHz. Reference: BLIMP paper
   unsigned m_blimpCoreBitWidth = m_GDLWidth;
+  unsigned m_simdUnitCount = m_blimpCoreBitWidth/32; // 32-bit SIMD unit
```
fasiddique, line 34:
```suggestion
  unsigned m_simdUnitCount = m_blimpCoreBitWidth / 32; // 32-bit SIMD unit
```

libpimeval/src/pimPerfEnergyBankLevel.cpp
deyuan: Yes, we need to run a bit-serial micro-program or an ALU instruction to convert data types

deyuan:
> > > > **Performance Per Watt** is defined as:
> > > > ```
> > > > Performance Per Watt = Total Operations / Total Energy (in Joules)
> > > > Total Ops = Vector Size × Number of Operation 
> > > > // For Pop Count, Number of Operation, Fulcrum = 12; Bank = 1; Bit-Serial = 1
> > > > // For Scaled Add, Number of Operation, Fulcrum = Bank = Bit-Serial = 2 
> > > > ```
> > > 
> > > 
> > > @deyuan please let me know if you have any concerns regarding bit-serial total op calculation formula
> > 
> > 
> > Do you think if we can treat popcount per element as one OP for Fulcrum? Those 12 operations per popcount on Fulcrum are similar like bit-serial micro-programs IMO, which are internal details.
> 
> This is a good question, and I also considered treating this as a single instruction for Fulcrum. However, the Fulcrum ALU does not natively support popcount instruction. As a result, I treated popcount as a fused operation like `scaled add`. @alifahmed, as you were involved with the fulcrum work, could you please let us know if we should consider fulcrum pop count as an stand alone instruction or a fused operation

I think we have the flexibility to extend the Fulcrum ALU to support new instructions, including popcount, bit extraction/insertion, type conversion, etc.

libpimeval/src/pimPerfEnergyBitSerial.cpp
deyuan: If we are focusing on computational OPs, then COPY_O2O can be ignored.
I think bit slice insertion and extraction can be treated as single bitwise computational OPs, and they are also supported by Fulcrum and bank-level.

fasiddique:
> > > > > **Performance Per Watt** is defined as:
> > > > > ```
> > > > > Performance Per Watt = Total Operations / Total Energy (in Joules)
> > > > > Total Ops = Vector Size × Number of Operation 
> > > > > // For Pop Count, Number of Operation, Fulcrum = 12; Bank = 1; Bit-Serial = 1
> > > > > // For Scaled Add, Number of Operation, Fulcrum = Bank = Bit-Serial = 2 
> > > > > ```
> > > > 
> > > > 
> > > > @deyuan please let me know if you have any concerns regarding bit-serial total op calculation formula
> > > 
> > > 
> > > Do you think if we can treat popcount per element as one OP for Fulcrum? Those 12 operations per popcount on Fulcrum are similar like bit-serial micro-programs IMO, which are internal details.
> > 
> > 
> > This is a good question, and I also considered treating this as a single instruction for Fulcrum. However, the Fulcrum ALU does not natively support popcount instruction. As a result, I treated popcount as a fused operation like `scaled add`. @alifahmed, as you were involved with the fulcrum work, could you please let us know if we should consider fulcrum pop count as an stand alone instruction or a fused operation
> 
> I think we have the flexibility to extend the Fulcrum ALU to support new instructions, including popcount, bit extraction/insertion, type conversion, etc.

Yes, this is true for any architecture. However, supporting those instructions would require a new ALU design, which would impact both area and energy modeling. If we introduce these instructions as standalone operations rather than fused instructions, it would no longer align with the original Fulcrum model. Instead, it would represent a modified version of Fulcrum, requiring us to obtain updated area and energy estimates using CACTI. I think this goes beyond the scope of the current paper, so I want to be as precise as possible in our modeling. Please correct me if I am misunderstanding something.

alifahmed:
> > > > **Performance Per Watt** is defined as:
> > > > ```
> > > > Performance Per Watt = Total Operations / Total Energy (in Joules)
> > > > Total Ops = Vector Size × Number of Operation 
> > > > // For Pop Count, Number of Operation, Fulcrum = 12; Bank = 1; Bit-Serial = 1
> > > > // For Scaled Add, Number of Operation, Fulcrum = Bank = Bit-Serial = 2 
> > > > ```
> > > 
> > > 
> > > @deyuan please let me know if you have any concerns regarding bit-serial total op calculation formula
> > 
> > 
> > Do you think if we can treat popcount per element as one OP for Fulcrum? Those 12 operations per popcount on Fulcrum are similar like bit-serial micro-programs IMO, which are internal details.
> 
> This is a good question, and I also considered treating this as a single instruction for Fulcrum. However, the Fulcrum ALU does not natively support popcount instruction. As a result, I treated popcount as a fused operation like `scaled add`. @alifahmed, as you were involved with the fulcrum work, could you please let us know if we should consider fulcrum pop count as an stand alone instruction or a fused operation

For the original fulcrum, pop count should be a fused operation. But that can always be extended to a single instruction with the overhead of updating area and energy cost.

libpimeval/src/pimPerfEnergyAquabolt.cpp
alifahmed, line 1: Just a reminder that, Aquabolt needs to be updated for read, write and logic percentage as well.

libpimeval/src/pimPerfEnergyFulcrum.cpp
```diff
@@ -80,6 +83,7 @@ pimPerfEnergyFulcrum::getPerfEnergyForFunc1(PimCmdEnum cmdType, const pimObjInfo
       energyLogical += ((m_eAP * 2) + (minElementPerRegion - 1) * 2 *  m_fulcrumShiftEnergy) + (minElementPerRegion * m_fulcrumALULogicalEnergy * numberOfALUOperationPerElement);
       mjEnergy = energyLogical * numCores;
       mjEnergy += m_pBChip * m_numChipsPerRank * m_numRanks * msRuntime;
+      totalOp = obj.getNumElements();
```
alifahmed, line 86: If `bit slice extract` is being done using one ALU op, than this is correct (which seems to be the case from the code).

libpimeval/src/pimPerfEnergyAquabolt.cpp
fasiddique: Updated.

fasiddique:
Thanks @alifahmed @deyuan for the feedback. I have addressed the comments and updated the tests as well.

deyuan (approved):

PR #265, opened 2025-03-31
[TACO] ResNet18 Implementation (PIM+CPU+GPU)

hugoabbot:
First draft of ResNet18 end-to-end implementation based off of VGGs.

### DIFF
--- a/PIMbench/resnet18/PIM/resnet18.cpp
+++ b/PIMbench/resnet18/PIM/resnet18.cpp
@@ -1,0 +1,963 @@
+// Test: C++ version of ResNet18
+// Copyright (c) 2024 University of Virginia
+// This file is licensed under the MIT License.
+// See the LICENSE file in the root of this repository for more details.
+
+#include "libpimeval.h"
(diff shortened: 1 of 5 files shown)


PIMbench/resnet18/PIM/resnet18.cpp
```diff
@@ -0,0 +1,963 @@
...
+    default:
+      fprintf(stderr, "\nUnrecognized option!\n");
+      usage();
+      exit(0);
+    }
+  }
+  return p;
+}
+
+void convolutionBlock(std::vector<std::vector<std::vector<int>>> &inputMatrix) {
```
fasiddique, line 90: is this an empty function?

PIMbench/resnet18/PIM/resnet18.cpp
```diff
@@ -0,0 +1,963 @@
...
+  return;
+}
+
+void identityBlock() {
```
fasiddique, line 94: This also seems like an empty function

PR #266, opened 2025-04-01
[PIMbench] Improve Image Downsampling Performance for GPU

Arleee1:
- Speedup image downsampling GPU kernel by reducing block.x
- Minor cleanup


fasiddique (approved):
Thank you!

PR #267, opened 2025-04-01
Convolution optimized

fasiddique:
@hugoabbot this convolution does not require host execution. I have tested for a few input but it would be good to test for resnet inputs.

### DIFF
--- a/PIMbench/convolution/PIM/conv.cpp
+++ b/PIMbench/convolution/PIM/conv.cpp
@@ -4,24 +4,24 @@
 // See the LICENSE file in the root of this repository for more details.
 
 #include "libpimeval.h"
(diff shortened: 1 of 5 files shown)


fasiddique:
PS This can be further optimized using tree-based addition instead of sequential addition O(m) vs O(log(m)). I do not have time now to work on this right away, but if you are interested, we can discuss the algorithm. Well, to be fair, I tried to implement that one first, but there was a minor bug on my implementation that I could not figure out. As a result, there's a mismatch on the CPU and PIM. 🤷🏽‍♀️

fasiddique:
Needs to be reviewed and merged. I am working on porting these changes for batched version in a separate PR. Thank you!

PR #277, opened 2025-05-02
Integration of AES S-box API and Refinement of GPU Power and Energy Measurement Methodology

mhgholamrezaei:
Changes:
- Introduced a dedicated API for the AES S-box operation in the PIMeval header.
- Added a new PIM command by extending the PimCmdFunc1 class to support S-box functionality.
- Implemented performance and energy estimation for the AES S-box on bit-serial architectures in pimPerfEnergyBitSerial.cpp.
- Implemented performance and energy estimation for the AES S-box on bank-level architectures in pimPerfEnergyBankLevel.cpp.
- Updated the AES benchmark to utilize the newly introduced S-box API.
- Refined the methodology for measuring GPU time, power, and energy to improve accuracy and consistency.


libpimeval/src/pimPerfEnergyBankLevel.cpp
```diff
@@ -326,3 +326,52 @@ pimPerfEnergyBankLevel::getPerfEnergyForRotate(PimCmdEnum cmdType, const pimObjI
...
+  double msRead = 0.0;
+  double msWrite = 0.0;
+  double msCompute = 0.0;
+  unsigned numPass = obj.getMaxNumRegionsPerCore();
+  unsigned bitsPerElement = obj.getBitsPerElement(PimBitWidth::ACTUAL);
+  uint64_t totalOp = 0;
+  unsigned numCores = obj.isLoadBalanced() ? obj.getNumCoreAvailable() : obj.getNumCoresUsed();
+
+  unsigned maxElementsPerRegion = obj.getMaxElementsPerRegion();
+  double numberOfOperationPerElement = 1;  // Assuming each 8-bit element allocates one ALU word
```
fasiddique, line 344: I think it would be better to not hard code this.

fasiddique:
Thanks @mhgholamrezaei! I have some questions:
1. Why the API was not modeled for Fulcrum?
2. For Bank-level, I can see that write latency is being added to the runtime. Could you please elaborate your reasoning behind the performance and energy modeling for the bank-level PIM.

mhgholamrezaei:
@fasiddique, thanks for the review. 

1. The lookup table (LUT) is not implemented for the Fulcrum architecture due to the high overhead of maintaining a LUT per core. Per Kevin’s suggestion, the LUT is only supported in the Bank-level architecture, and the S-box operation is offloaded to the host during Fulcrum execution.
2. The performance and energy model for the Bank-level architecture is based on the Func1 operation, as the S-box involves a single input and a single output. Additionally, it is assumed that the LUT access latency is absorbed within the pipeline.

fasiddique:
> @fasiddique, thanks for the review.
> 
> 1. The lookup table (LUT) is not implemented for the Fulcrum architecture due to the high overhead of maintaining a LUT per core. Per Kevin’s suggestion, the LUT is only supported in the Bank-level architecture, and the S-box operation is offloaded to the host during Fulcrum execution.
Understood! But even if Fulcrum does not support this, we need to add a switch case for this in the Fulcrum, may be in the func1 command. So that, if the user run this benchmark for Fulcrum, they will not get some segfault or any other error.
> 2. The performance and energy model for the Bank-level architecture is based on the Func1 operation, as the S-box involves a single input and a single output. Additionally, it is assumed that the LUT access latency is absorbed within the pipeline.
Thank you! I am thinking in this case can we just use the func1 and add switch cases to it for the new API?

mhgholamrezaei:
@fasiddique, thanks for your suggestions.

1. For the Fulcrum performance model, we can display an error message and halt execution. Should this check be implemented within the PIM command itself, or would it be better to create a new performance model class and handle the error there? Do you know of any existing command where the performance model is not implemented for a specific architecture, as a reference?

2. The performance behavior of this command is similar to Func1, but it also takes an additional parameter lut. Integrating this into Func1 would reduce code readability and introduce added complexity—especially within the sanityCheck function. I would prefer to implement a separate PIM command for the S-box. However, if you believe it’s more appropriate to merge it into Func1, I’m open to that as well.

fasiddique:
> @fasiddique, thanks for your suggestions.
> 
> 1. For the Fulcrum performance model, we can display an error message and halt execution. Should this check be implemented within the PIM command itself, or would it be better to create a new performance model class and handle the error there? Do you know of any existing command where the performance model is not implemented for a specific architecture, as a reference?
> 2. The performance behavior of this command is similar to Func1, but it also takes an additional parameter lut. Integrating this into Func1 would reduce code readability and introduce added complexity—especially within the sanityCheck function. I would prefer to implement a separate PIM command for the S-box. However, if you believe it’s more appropriate to merge it into Func1, I’m open to that as well.

Some suggestions:
I went over the code and saw that this API can be handled by Func1 or Func2 performance/energy model. There's no need to create a separate performance/energy modeling function for this for different PIM architecture. The reason being, this instruction does not require any special handling. We needed to separate reduction, broadcast and rotation because they needed some special handling like inter-bank communication, host execution and such. In the command class, however, you can have separate class for this. 
If an architecture does not support an API, you can show something like following, under that commands switch case.
`std::cout << "PIM-Warning: Unsupported for Fulcrum: " << pimCmd::getName(cmdType, "") << std::endl;`
I am also interested to know, if the user want to run AES benchmark for Fulcrum, how will they go about it? Do we have separate implementation for Fulcrum vs other PIM architectures?

mhgholamrezaei:
Thank you!

`pimCmdFunc1` is exteded in the new version for implementation of the AES functions. 

For the Fulcrum architecture, AES execution is handled by the host. The `aesSubBytes` function in aes.cpp will check the PIM architecture, and if it is Fulcrum, the S-box computation will be performed on the host.

libpimeval/src/pimCmd.h
```diff
@@ -266,6 +271,10 @@ class pimCmdFunc1 : public pimCmd
         }
         break;
     }
+    case PimCmdEnum::AES_SBOX:
+    case PimCmdEnum::AES_INVERSE_SBOX:
+      result = m_lut[operand];
```
deyuan, line 276: For inverse s-box, do you expect we do inverse lookup or just pass in the inverse lut? If with the latter, a single AES_SBOX command would be sufficient

libpimeval/src/pimPerfEnergyAquabolt.h
```diff
@@ -27,7 +27,7 @@ class pimPerfEnergyAquabolt : public pimPerfEnergyBase
   virtual pimeval::perfEnergy getPerfEnergyForReduction(PimCmdEnum cmdType, const pimObjInfo& obj, unsigned numPass) const override;
   virtual pimeval::perfEnergy getPerfEnergyForBroadcast(PimCmdEnum cmdType, const pimObjInfo& obj) const override;
   virtual pimeval::perfEnergy getPerfEnergyForRotate(PimCmdEnum cmdType, const pimObjInfo& obj) const override;
-
+
```
deyuan, line 30: Could you revert some line touching with trailing spaces?

libpimeval/src/libpimeval.h
```diff
@@ -9,7 +9,7 @@
 
 #include <cstdint>
 #include <cstdarg>
-
+#include <vector>
```
deyuan, line 12: Need to keep an empty line

libpimeval/src/pimPerfEnergyTables.cpp
```diff
@@ -211,6 +211,8 @@ pimPerfEnergyTables::bitsimdPerfTable = {
       { PimCmdEnum::REDMIN_RANGE, {   32,    8,   42 } },
       { PimCmdEnum::REDMAX,       {   32,    8,   42 } },
       { PimCmdEnum::REDMAX_RANGE, {   32,    8,   42 } },
+      { PimCmdEnum::AES_SBOX,     {   2450, 2112, 3625 } },
+      { PimCmdEnum::AES_INVERSE_SBOX,     {   2420, 2086, 3610 } },
```
fasiddique, line 215: Has this been updated based on the discussion?

libpimeval/src/pimPerfEnergyTables.cpp
```diff
@@ -531,6 +533,8 @@ pimPerfEnergyTables::bitsimdPerfTable = {
       { PimCmdEnum::REDMIN_RANGE, {   32,    8,   51 } },
       { PimCmdEnum::REDMAX,       {   32,    8,   51 } },
       { PimCmdEnum::REDMAX_RANGE, {   32,    8,   51 } },
+      { PimCmdEnum::AES_SBOX,     {   2450, 2112, 3625 } },
+      { PimCmdEnum::AES_INVERSE_SBOX,     {   2420, 2086, 3610 } },
```
fasiddique, line 537: Has this been updated based on the discussion?

libpimeval/src/pimCmd.h
mhgholamrezaei: Since the bit-serial architecture has different performance models for the forward and inverse sbox functions, it is essential to have two separate commands for them.

libpimeval/src/pimPerfEnergyTables.cpp
mhgholamrezaei: Yes. The stats are based on the `inv_and_xnor_mux` architecture.

libpimeval/src/pimPerfEnergyTables.cpp
mhgholamrezaei: Yes.

libpimeval/src/pimPerfEnergyBankLevel.cpp
```diff
@@ -80,6 +80,11 @@ pimPerfEnergyBankLevel::getPerfEnergyForFunc1(PimCmdEnum cmdType, const pimObjIn
       totalOp = obj.getNumElements();
       break;
     }
+    case PimCmdEnum::AES_SBOX:
+    case PimCmdEnum::AES_INVERSE_SBOX:
+    {
+      numberOfOperationPerElement = 1;  // Assuming each 8-bit element allocates one ALU word
```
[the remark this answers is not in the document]
mhgholamrezaei: @fasiddique, could you clarify what object you're referring to in the comment "TODO: Use object to decide the value of numberOfOperationPerElement"?

libpimeval/src/pimCmd.h
deyuan: Makes sense. Thanks!

libpimeval/src/libpimeval.h
```diff
@@ -198,6 +198,11 @@ PimStatus pimShiftElementsLeft(PimObjId src);
 PimStatus pimShiftBitsRight(PimObjId src, PimObjId dest, unsigned shiftAmount);
 PimStatus pimShiftBitsLeft(PimObjId src, PimObjId dest, unsigned shiftAmount);
 
+// AES sbox and inverse-box APIs
+// Note: AES S-box and inverse S-box are treated separately because their bit-serial performance models differ.
+// However, it is the user's responsibility to provide the appropriate LUT to ensure correct functionality.
```
deyuan, line 203: It would be great to mention that pimAesInverseSbox requires an inverse LUT as the input parameter

deyuan (approved):

tests/test-aes-sbox/test-aes-sbox.cpp
```diff
@@ -0,0 +1,116 @@
...
+  PimObjId srcObj = pimAlloc(PIM_ALLOC_AUTO, numElements, PIM_UINT8);
+  assert(srcObj != -1);
+
+  PimObjId destObj = pimAllocAssociated(srcObj, PIM_UINT8);
+  assert(destObj != -1);
+
+  status = pimCopyHostToDevice((void*)src.data(), srcObj);
+  assert(status == PIM_OK);
+
+  status = pimAesSbox(srcObj, destObj, lut_vec);
```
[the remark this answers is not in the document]
mhgholamrezaei: Thanks. It is done.

libpimeval/src/libpimeval.h
mhgholamrezaei: Thanks for your comment. It is addressed in the new commit.

PR #278, opened 2025-05-05
Chunked Stencil Implementation PIM

Arleee1:
- Iterative Stencil implementation for PIM, including support for chunking the operation within PIM regions
- Add default parameter for enabling cross region shifts in pimShiftElementsLeft/Right
- TODO: Handle load balanced fulcrum


libpimeval/src/pimCmd.h
```diff
@@ -506,8 +506,8 @@ class pimCmdBroadcast : public pimCmd
 class pimCmdRotate : public pimCmd
 {
 public:
-  pimCmdRotate(PimCmdEnum cmdType, PimObjId src)
-    : pimCmd(cmdType), m_src(src)
+  pimCmdRotate(PimCmdEnum cmdType, PimObjId src, bool useCrossRegionCommunication)
```
fasiddique, line 509: Does this work for rotate API? I haven't seen that rotate API in the PIMSim class was modified to incorporate this argument.

Arleee1: Currently it only works for pimShiftElements,not pimRotateElements,as stencil only uses the former, and the API is temporary,but it should be easily extendable to both if you would like

fasiddique: That's fine!

PR #285, opened 2025-05-18
updated prefixsum

Beenishgul:
Please the updated version.

### DIFF
--- a/misc-bench/cpp-prefix-sum/prefix-sum.cpp
+++ b/misc-bench/cpp-prefix-sum/prefix-sum.cpp
@@ -7,11 +7,15 @@
 #include <vector>
 #include <cmath>
 #include <algorithm>
-#include "util.h"
+#include "../../util/util.h"
 #include "libpimeval.h"
 #include <getopt.h>
 #include <stdint.h>
 #include <iomanip>
+#include <chrono>
+
+std::chrono::duration<double, std::milli> hostElapsedTime = std::chrono::duration<double, std::milli>::zero();
+
 
 #if defined(_OPENMP)
 #include <omp.h>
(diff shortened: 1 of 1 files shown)


misc-bench/cpp-prefix-sum/prefix-sum.cpp
```diff
@@ -77,138 +81,263 @@ struct Params getInputParams(int argc, char **argv)
...
+    for (uint64_t i = 0; i < input.size(); i++) {
+        deviceoutput.push_back(0);
+    }
 
-  // Verification of Results hostresults vs deviceresults
-  if (params.shouldVerify)
-  {
+    hostoutput[0] = input[0];
+    for (uint64_t i = 0; i < input.size(); i++) {
+        hostoutput[i + 1] = hostoutput[i] + input[i + 1];
```
fasiddique, line 227: This is an out-of-bounds access. Depending on the compiler and operating system, it may either get segmentation fault, or produce error-prone results.

fasiddique:
```
prefix-sum.out(9981,0x1f217cf40) malloc: Incorrect checksum for freed object 0x12200c800: probably modified after being freed.

Corrupt value: 0xffc00

prefix-sum.out(9981,0x1f217cf40) malloc: *** set a breakpoint in malloc_error_break to debug

Abort trap: 6
```

misc-bench/cpp-prefix-sum/prefix-sum.cpp
```diff
@@ -77,138 +81,263 @@ struct Params getInputParams(int argc, char **argv)
...
+    // UpSweep
+    while (intermeadiate_results.size() > 1) {
+        std::vector<int> even, odd;
+
+        for (uint64_t i = 0; i < intermeadiate_results.size(); ++i) {
+            if (i % 2 == 0)
+                even.push_back(intermeadiate_results[i]);
+            else
+                odd.push_back(intermeadiate_results[i]);
+        }
```
fasiddique, line 244: It's recommended to avoid using `push_back` for efficiency purpose, because this allocates memory dynamically. Since the vector size is known beforehand, we can use `reside` and indexing.

fasiddique: *`resize`

misc-bench/cpp-prefix-sum/prefix-sum.cpp
```diff
@@ -77,138 +81,263 @@ struct Params getInputParams(int argc, char **argv)
...
+    }
 
-  // Verification of Results hostresults vs deviceresults
-  if (params.shouldVerify)
-  {
+    hostoutput[0] = input[0];
+    for (uint64_t i = 0; i < input.size(); i++) {
+        hostoutput[i + 1] = hostoutput[i] + input[i + 1];
+        intermeadiate_results.push_back(input[i]);
+        host_device_merged.push_back(input[i]);
```
fasiddique, line 229: Why do we need to have three copies of the input vector?

PR #286, opened 2025-05-19
Prefix sum API

fasiddique:


fasiddique:
@alifahmed For total OP calculation in prefix sum, should we consider `operation count = 2 * num of elements`?

alifahmed:
> @alifahmed For total OP calculation in prefix sum, should we consider `operation count = 2 * num of elements`?

Yes, because, this is doing two addition per element.

alifahmed (approved):

PR #287, opened 2025-05-20
Pim+host prefixsum

Beenishgul:


misc-bench/Makefile
```diff
@@ -1,8 +1,42 @@
+<<<<<<< HEAD
 # Makefile for PIMeval Simulator - Applications
+=======
+<<<<<<< HEAD
+# Makefile for PIMeval / PIMbench Framework
+=======
+# Makefile for PIMeval Simulator - Applications
+>>>>>>> 556bf46 (host_PIM_Prefixsum)
+>>>>>>> 14b1d2b (host_PIM_Prefixsum)
```
fasiddique, line 9: Looks like there's unresolved merge conflict at the top of the Makefile. The conflict markers (`<<<<<<<`, `=======`, `>>>>>>>`) need to be removed and the appropriate header should be chosen.
Please resolve the conflict.

misc-bench/Makefile
fasiddique, line 1: This entire file contains unresolved merge conflicts with markers. This needs to be cleaned up.

PR #289, opened 2025-05-21
Support up to 16 regs for digital bit-serial PIM.

deyuan:
This is for experimental purposes and for micro-ops simulation only.
This has no impact on functional simulation and can be merged directly.


mhgholamrezaei (approved):
It looks good to me!

PR #318, opened 2025-11-05
Fix make command from root

Arleee1:
Previously, `make debug` from the root directory failed as the string matching data generators did not have this Makefile target. This PR adds the targets `debug`, `perf`, and `dramsim3_integ`, to both data generator Makefiles, sets them to make the data generator, and declares them as phony.


fasiddique (approved):
Thanks!

PR #322, opened 2026-02-25
BFS, Shift-element, Local Indexing

fasiddique:


PIMbench/bfs/PIM/bfs.cpp
```diff
@@ -0,0 +1,643 @@
+// Test: C++ version of breadth-first search
+// Copyright (c) 2024 University of Virginia
```
Arleee1: [nit] Should this be 2026?

libpimeval/src/libpimeval.h
```diff
@@ -156,40 +167,43 @@ PimStatus pimCopyObjectToObject(PimObjId src, PimObjId dest);
 PimStatus pimConvertType(PimObjId src, PimObjId dest);
 
 // Logic and Arithmetic Operation
+// Note: idxBegin and idxEnd specify the range [idxBegin, idxEnd) of indexes to be processed by the PIM.
+// If the default values (OULL) for idxBegin and idxEnd are used, the entire range of the PIM object will be considered.
+// TODO: Currently idxBegin and idxEnd for all PIM Object IDs are same; however, it is possible to perform operation on
```
fasiddique:
```suggestion
// TODO: Currently idxBegin and idxEnd for all PIM Object IDs are same; however, it is possible to perform operation on  same range  of different objects but different idxBegin and idxEnd
```

